PR #10, opened 2015-06-09
fix date parsing and typo

mgwilliams:
Signed-off-by: Matthew Williams <mgwilliams@gmail.com>

Although the API spec says unix timestamps are returned by the API, in fact they are of the format `2015-06-09T21:05:29.00502626Z` (this is true in 0.9 through 0.11).

If it is desirable to avoid the dependency on python-dateutil, I can write a helper function using strptime. Or pylxd could just return the timestamp verbatim as returned by the API, since it's already human readable.

This PR also fixes a small typo that was causing a `404`.

### DIFF
--- a/pylxd/operation.py
+++ b/pylxd/operation.py
@@ -13,6 +13,7 @@
 #    under the License.
 
 import datetime
+from dateutil.parser import parse as parse_date
 
 from . import connection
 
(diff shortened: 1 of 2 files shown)


tych0:
On Tue, Jun 09, 2015 at 03:12:31PM -0700, Matthew Williams wrote:
> Signed-off-by: Matthew Williams <mgwilliams@gmail.com>
> 
> Although the API spec says unix timestamps are returned by the API, in fact they are of the format `2015-06-09T21:05:29.00502626Z` (this is true in 0.9 through 0.11).

Thanks for pointing this out. I like the string format myself, so I've
filed:

https://github.com/lxc/lxd/pull/752

Tycho

techtonik:
So, what is the current output format and proposed?

PR #16, opened 2015-07-02
Fix PEP8 errors

Saviq:


pylxd/container.py
```diff
@@ -14,34 +13,29 @@
...
+
 
 class LXDContainer(base.LXDBase):
     # containers:
+
     def container_list(self):
         (state, data) = self.connection.get_object('GET', '/1.0/containers')
-        return [container.split('/1.0/containers/')[-1] for container in data['metadata']]
-
-    def container_defined(self, container):
```
Saviq: This method was defined twice, hence deleted. Real implementation below (line 49/43).

PR #37, opened 2015-10-27
faster behavior for determining if a container is defined

tych0:
Instead of trying to get the state (which causes LXD to do a bunch of
forking/setnsing crap that is quite slow). Instead, let's just query the
list of the containers, which, in LXD is a DB query, and check if our
container is in that list.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -15,6 +15,7 @@
 import json
 
 from pylxd import base
+from pylxd import exceptions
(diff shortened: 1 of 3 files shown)


tych0:
(mostly just for james and I testing, no idea if this is actually a good idea or not)

PR #38, opened 2015-10-27
use a while True to wait instead of LXD's API

tych0:
Here's a patch that might not be so nice, but we think this is the easiest
way to do it for the time being.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>


tych0:
Closing this, as it wasn't the source of the bug.

Issue #26, opened 2015-08-26
Feature: operations websocket


aarnaud:
Hi,

i will wish use this API : /1.0/operations/<uuid>/websocket

Thanks,

uglide:
:+1:

tych0:
On Wed, Aug 26, 2015 at 07:53:47AM -0700, Anthony Arnaud wrote:
> Hi,
> 
> i will wish use this API : /1.0/operations/<uuid>/websocket

Isn't this implemented by operation_stream()?

uglide:
@tych0 It's not implemented by operation_stream(). [get_ws() function in connection module is not implemented properly](https://github.com/lxc/pylxd/blob/master/pylxd/connection.py#L130-L134)

Current implementation returns only status code instead of web socket object.

uglide:
This issue should be closed because appropriate pull request was merged (#41)

PR #41, opened 2015-11-05
Remove get_ws() method stub with real websocket implementation

uglide:
This commit adds websockets support in PyLXD library which allows
users track progress and output of executed operations and commands.

This commit is required to use PyLXD in Openstack Manila project.

This pull request closes issue #26

### DIFF
--- a/pylxd/connection.py
+++ b/pylxd/connection.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2015 Canonical Ltd
+# Copyright (c) 2015 Mirantis inc.
 #
 #    Licensed under the Apache License, Version 2.0 (the "License"); you may
 #    not use this file except in compliance with the License. You may obtain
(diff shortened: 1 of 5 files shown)


javacruft:
Hi Igor

Thanks for the patch - in review now; what are the plans for use of PyLXD in Manila?  Would be interested in understanding the use case...

uglide:
Hello James

Currently we do investigation around LXD. Our goal is creation of the driver for Manila which uses LXC/LXD for tenant isolation and LVM on bare metal host as storage.

zulcss:
pylxd is no necessarily an openstack project can we make eventlet optional?

uglide:
Yes, we can. So do you want to have support for geven and eventlet? Or it should be possible to run pylxd without both?

zulcss:
It should be possible to run pylxd without both.

uglide:
@zulcss @javacruft Updated version was uploaded. Hopefully, it can be merged soon.

zulcss:
thanks looks good

Issue #42, opened 2015-11-13
Test failures on Ubuntu 14.04


javacruft:
```
==============================
Failed 2 tests - output below:
==============================

pylxd.tests.test_connection.LXDInitConnectionTest.test_https_connection
-----------------------------------------------------------------------

Captured traceback:
    Traceback (most recent call last):
      File "/usr/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
        return func(*args, **keywargs)
      File "/��PKGBUILDDIR��/pylxd/tests/test_connection.py", line 48, in test_https_connection
        conn.connect()
      File "/��PKGBUILDDIR��/pylxd/connection.py", line 57, in connect
        ssl_version=ssl.PROTOCOL_TLSv1_2)
    AttributeError: 'module' object has no attribute 'PROTOCOL_TLSv1_2'
    
pylxd.tests.test_connection.LXDInitConnectionTest.test_https_proxy_connection
-----------------------------------------------------------------------------
Captured traceback:
    Traceback (most recent call last):
      File "/usr/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
        return func(*args, **keywargs)
      File "/��PKGBUILDDIR��/pylxd/tests/test_connection.py", line 66, in test_https_proxy_connection
        conn.connect()
      File "/��PKGBUILDDIR��/pylxd/connection.py", line 57, in connect
        ssl_version=ssl.PROTOCOL_TLSv1_2)
    AttributeError: 'module' object has no attribute 'PROTOCOL_TLSv1_2'
```

javacruft:
Updated in repo; closing issue.

PR #47, opened 2015-11-19
Detect SSL TLS version

zulcss:
Ubuntu 14.04 doesnt have ssl.PROTOCOL_TLSv1_2, so check  to see if we have ssl.PROTOCOL_TLSv1_2 available otherwise use OpenSSL.SSL.TLSv1_2_METHOD.

Fixes issue #42

### DIFF
--- a/pylxd/connection.py
+++ b/pylxd/connection.py
@@ -22,6 +22,22 @@ from pylxd import exceptions
 from pylxd import utils
 from six.moves import http_client
 
+if hasattr(ssl, 'SSLContext'):
+    # For Python >= 2.7.9 and Python 3.x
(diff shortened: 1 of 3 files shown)


pylxd/connection.py
```diff
@@ -22,6 +22,22 @@
 from pylxd import utils
 from six.moves import http_client
 
+if hasattr(ssl, 'SSLContext'):
```
rockstar: I think it'd be better to try/except here, rather than the if. It's semantic, so not really a big deal, but a more "pythonic" solution.

rockstar:
Just a small suggestion in the patch, but otherwise, :+1: to merge.

PR #51, opened 2015-12-14
Remove override of close() in WebSocketClient

uglide:
Since 0.24 LXD doesn't terminate web socket and it should be closed by client via close() method.


zulcss:
lgtm thanks

PR #54, opened 2016-01-13
The end is near

rockstar:
This is the start of a re-work of the LXD api client for python.

The next step, after this, is to go through the LXD spec itself and implement the client per the spec, thus making sure the client is spec'd properly, but also forcing LXD to have implemented the spec properly.

This code isn't currently being used yet. Once the full client is implemented, we'll deprecate the old code (which, we've recently learned, doesn't work at all on python3).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "2.7"
   - "3.4"
+  - "pypy"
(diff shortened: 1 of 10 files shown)


zulcss:
Can you fix your pep8 errors, otherwise i would be +1

PR #59, opened 2016-02-16
2 0 api

rockstar:
This is the first hit at the pylxd 2.0 api.  It's simple, and it doesn't catch many errors, but it also doesn't overcatch errors like the current version of pylxd does.  The API is pretty simple:

    >>> lxd = pylxd.client.Client()
    >>> container = lxd.containers.get('an-container')
    >>> container.restart()

There are some things about snapshots that still need to get worked out before `nova-lxd` can completely cut over, but with @zulcss' work on the deploy-from-master stuff, that should be really quick to iterate on.

I will say that we will lose some features in the move to 2.0, but with the current state of our testing, I'm not sure a lot of those features work anyway.  I purposefully didn't remove any old code, and won't until right before we cut a pylxd release for 2.0 (matching the LXD 2.0 release).


zulcss:
We have dropped the websocket stuff correct? If yes can you drop ws4py from the dependencies as well.

zulcss:
+1 from me once the cosmetic stuff is cleaned up.

PR #64, opened 2016-03-01
Add state function on container to return state with network. 

aarnaud:
exemple: container.state().network['eth0']


zulcss:
lgtm

PR #65, opened 2016-03-01
Update version number for install to work

john5223:


aarnaud:
Version 2.0.0b2 already commit in https://github.com/lxc/pylxd/pull/64

PR #66, opened 2016-03-04
Fix container migration

zulcss:
Fix container migration for recent versions of LXD.


rockstar:
:+1: Merge it.

It does seem weird to be adding stuff to the deprecated api, but we've already talked about that, and will begin using the 2.0 API after 16.04.

PR #68, opened 2016-03-06
Fix integration test suite

zulcss:
Fix integration test suite.

### DIFF
--- a/integration/test_containers.py
+++ b/integration/test_containers.py
@@ -26,7 +26,6 @@ class TestContainers(IntegrationTestCase):
         self.assertEqual(name, container.name)
         self.addCleanup(self.delete_container, name)
 
-
     def test_all(self):
         """A list of all containers is returned."""
         name = self.create_container()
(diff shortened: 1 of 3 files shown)


integration/test_containers.py
```diff
@@ -20,11 +20,11 @@ class TestContainers(IntegrationTestCase):
     def test_get(self):
         """A container is fetched by name."""
         name = self.create_container()
-        self.addCleanup(self.delete_container, name)
 
         container = self.client.containers.get(name)
 
         self.assertEqual(name, container.name)
+        self.addCleanup(self.delete_container, name)
```
rockstar: Why did you move this? If the test fails, the cleanup task won't be added and thus won't be cleaned up.

integration/test_containers.py
```diff
@@ -84,13 +85,6 @@ def test_rename(self):
         container = self.client.containers.get(name)
         self.assertEqual(name, container.name)
 
-    def test_delete(self):
```
rockstar: Why did this get moved?

zulcss:
Merged addressed comments to Paul already.

PR #70, opened 2016-03-08
Fix six.PY34 into six.PY3

thomasgoirand:
This otherwise fail to build in Jessie, and I'm adding this patch to the Debian package.


rockstar:
👍  Weird, I don't even remember writing this code. I don't think I knew `six.PY34` even existed.

PR #74, opened 2016-04-23
Python3 fixes

stgraber:
I needed those two changes to get lxc-to-lxd working with the non-deprecated pylxd API.

I'm reasonably sure the the quote commit is properly backward compatible, I'm not so sure about the iteritems() one though, if it's not, we'll need a wrapper function which takes a dict and calls the right thing depending on python version.

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -70,7 +70,7 @@ class Container(mixin.Waitable, mixin.Marshallable):
 
     def __init__(self, **kwargs):
         super(Container, self).__init__()
-        for key, value in kwargs.iteritems():
+        for key, value in six.iteritems(kwargs):
             setattr(self, key, value)
 
     def reload(self):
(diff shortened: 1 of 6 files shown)


pylxd/container.py
```diff
@@ -70,7 +70,7 @@ def create(cls, client, config, wait=False):
 
     def __init__(self, **kwargs):
         super(Container, self).__init__()
-        for key, value in kwargs.iteritems():
+        for key, value in kwargs.items():
```
rockstar: Hm. This concerns me. I realize what's going on here, but it may be in our best interest to just `import six` and use `six.iteritems` or something equivalent.

stgraber: Yeah, that's why I left that comment in the pull request.

There are several similar iteritems() uses throughout your codebase. Since I don't use python2 I can't really make sure that the fix is correct (be that a straight .items() or using six).

stgraber:
Updated to use six, completely untested.

rockstar:
👍

PR #85, opened 2016-05-23
Mock LXD services

rockstar:
As we've seen significant uptick in the use of pylxd recently, and some demand for good testing, this patch adds unit tests for the newer API. It adds a mock service for requests, so the entire test suite can be run without a LXD instance at all. Currently, it puts the unit test coverage at 70% of the new API. While writing these tests, I also found a bug in the `LXD_DIR` logic of the client, and fixed that (something initial tests would have caught...)

This infrastructure is a start, and should make it easier for others to contribute without having to make up some random mock service system. It's definitely not the end of what it should be (and there's even an `XXX` comment to that effect in this patch).

While running these tests on various platforms (to make sure they run fine outside of a lxd environment), I found an issue with the default TLS version, where the two separate if statements had some holes in the logic. I _believe_ I fixed those on the platforms that I've tested on, though my gut tells me there still might be an issue with how that default TLS is calculated.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,5 +3,6 @@ branch = True
 source = pylxd
 
 [report]
-ignore-errors = True
-omit = pylxd/tests/*
+omit =
+    pylxd/tests/*
+    pylxd/deprecated/*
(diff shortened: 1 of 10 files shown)


zulcss:
lgtm

jpic:
Thank you so much

:guitar:

PR #87, opened 2016-05-23
Fix integration test

jpic:
The purpose of this PR is only to fix `tox -e integration` or `nosetests integration` if you have a busybox image in your lxd.

### DIFF
--- a/integration/test_containers.py
+++ b/integration/test_containers.py
@@ -40,7 +40,7 @@ class TestContainers(IntegrationTestCase):
         """Creates and returns a new container."""
         config = {
             'name': 'an-container',
-            'architecture': 2,
+            'architecture': '2',
             'profiles': ['default'],
             'ephemeral': True,
             'config': {'limits.cpu': '2'},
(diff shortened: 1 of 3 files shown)


rockstar:
Hi @jpic - I appreciate the patch and the work you're doing. However, I don't want to get into the habit of running the integration tests on a lxd instance that matters, i.e. in an environment where we couldn't just delete all the containers and all the images and all the profiles after every run. I certainly don't want to be the result of data loss because someone ran the integration tests against their production environment.

The true intent of the integration tests is mostly for the upstream lxd developers to run in their CI. I'd rather not turn the integration tests into a maintenance burden for the pylxd developers (as we're already resource constrained).

jpic:
I don't understand how we can prevent someone from doing something as stupid as running potentially destructive tests on a production environment, is that the reason the PR that makes these tests pass was closed ?

jpic:
> I'd rather not turn the integration tests into a maintenance burden for the pylxd developers (as we're already resource constrained).

Well if you're looking to extend the family I'd be glad to volunteer.

PR #90, opened 2016-05-23
Container creation failure handling

jpic:
This is a simple PoC of how we could handle container creation. It's tested in integration, so it requires integration tests to pass ( #87 does just that ). So please, close your eyes on any commit that is NOT 673845c in this PR :)

What do you think ? Would that approach be fine ? Perhaps we could have have unit tests in addition to integration tests too ?


jpic:
Rebasing and re-opening.

PR #93, opened 2016-05-23
Setup LXD before running integration tests

jpic:
The busybox image is necessary for test_containers.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,11 @@ commands = nosetests pylxd
 commands = flake8
 
 [testenv:integration]
-commands = nosetests integration
+passenv = HOME
+whitelist_externals = lxc
+commands =
+    lxc image import https://dl.stgraber.org/lxd --alias busybox
+    nosetests integration
 
 [flake8]
 # E123, E125 skipped as they are invalid PEP-8.


rockstar:
I'm unsure why this is needed. There is already a busybox image that gets created by the integration tests.

jpic:
Yes, it is created and then dropped, and never labeled busybox, in test_images.py

But for test_containers.py, it is expected to have an image labeled busybox on the lxd.

Perhaps that could be done in a setUpClass classmethod, which I've tried, this requires to make another client instance, in addition to the one created in setUp(), so I thought this would be sufficient and not make test code more complex.

rockstar:
I've never had this faillure before, so I'm not sure where it's coming from. If there's a misconfiguration somewhere, I'd like to have it documented before we go adding more required configuration.

Also, I'd prefer not to have any outside network traffic if we can help it. That's why there's that script for creating a busybox image (cargo culted heavily from lxd itself), rather than just going out and getting busybox.

jpic:
> I've never had this faillure before, so I'm not sure where it's coming from.

How does the output looks like when you run integration tests ? Don't worry, I bet I know each error it can output by now :D

> Also, I'd prefer not to have any outside network traffic if we can help it. That's why there's that script for creating a busybox image (cargo culted heavily from lxd itself), rather than just going out and getting busybox.

Agreed, alternative proposed in #100.

rockstar:
Closing in favor of #100

PR #95, opened 2016-05-23
Expect test to have created 1 container

jpic:
This enables running tests on a development machine.

NOT for production !!


jpic:
An alternative here would be to **prevent tests from running** if the lxd instance doesn't have 0 containers, to make it clear from the beginning that these tests must have exclusivity on an lxd instance with which you may not play against on your own.

rockstar:
I would prefer that we do the alternative here: check to see if any images or containers exist, and bail out.

rockstar:
Closing after opening issue #118

PR #96, opened 2016-05-23
Handle container creation failure

jpic:


jpic:
I see, this is outdated.

PR #98, opened 2016-05-23
Fix #97: improvements for container creation failure

### LINKED ISSUE #97: Improvement for container creation failure
Currently, a runtime exception is raised when lxd responds with a failure to create a container. In user code, this makes it impossible to route exception handling ie.:

```
try:
    client.containers.create(dict(name='foo'))
except ContainerCreationFailure:
    print "Something went creating the container"
except:
    print "Something unknown went wrong, but the container was created"
```

Also, the runtime error doesn't carry the response, making it impossible for user code to use the HTTP response, ie. to display it:

```
try:
    client.containers.create(dict(name='foo'))
except ContainerCreationError as e:
    print "Server did not create the container, details: ", e.response.json()
```

Please let me know your thought about this, thanks again for the amazing job you're doing hosting this sprint !

jpic:
Enables routing raised module-specific exceptions based on their type:

```
try:
    client.containers.create(dict(name='foo'))
except ContainerCreationFailure:
    print "Something went creating the container"
except:
    print "Something unknown went wrong, but the container was created"
```

And also stuff with the http response such as:

```
try:
    client.containers.create(dict(name='foo'))
except ContainerCreationError as e:
    print "Server did not create the container, details: ", e.response.json()
```

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -31,3 +31,9 @@ Profile
 
 .. autoclass:: pylxd.profile.Profile
    :members:
+
+Exceptions
+----------
+
+.. automodule:: pylxd.exceptions
+   :members:
--- a/integration/test_containers.py
+++ b/integration/test_containers.py
@@ -13,6 +13,8 @@
 #    under the License.
 from integration.testing import IntegrationTestCase
 
+from pylxd.exceptions import ContainerCreationFailure
+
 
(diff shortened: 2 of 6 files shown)


rockstar:
Thanks for this patch. Unfortunately, we already had some work in flight for this same thing, in #103.

PR #100, opened 2016-05-24
Create busybox image at first create_container call

jpic:
Alternative to #93

### DIFF
--- a/integration/testing.py
+++ b/integration/testing.py
@@ -34,6 +34,20 @@ class IntegrationTestCase(unittest.TestCase):
 
     def create_container(self):
         """Create a container in lxd."""
+        if not getattr(type(self), '_busybox_ready', False):
+            path, fingerprint = create_busybox_image()
+
+            with open(path) as f:
+                self.client.images.create(f.read(), wait=True)
+
+            self.lxd.images.aliases.post(json={
+                'description': '',
+                'target': fingerprint,
+                'name': 'busybox',
+            })
+
+            type(self)._busybox_ready = True
+
         name = self.generate_object_name()
         machine = {
             'name': name,


integration/testing.py
```diff
@@ -34,6 +34,20 @@ def generate_object_name(self):
...
+                self.client.images.create(f.read(), wait=True)
+
+            self.lxd.images.aliases.post(json={
+                'description': '',
+                'target': fingerprint,
+                'name': 'busybox',
+            })
+
+            type(self)._busybox_ready = True
+
```
jpic: See line 59 of this file how it expects the 'busybox' image to be there.
2016-05-24-204543_537x198_scrot

rockstar:
I'm +1 on this patch, with the caveat that I'd like the image to be removed at the end of the test (i.e. using `addCleanup`). I will merge once that's in place.

jpic:
Should we do that in tearDownClass with a new Client instance, or should we reuse the test case instance client instance in tearDown and in this case find out if we're tearing-down the last test of the case and if so delete the image ?

rockstar:
I don't think you need to use `tearDown` at all. You can just use `addCleanup`, and only when you end up creating the busybox image.

While I think it's good to have a busybox image, the integration tests really do need to be run against a throwaway lxd instance. For example, a lxd instance running inside a container in your real lxd instance.

jpic:
In this patch, the busybox label is created once for the complete test
suite. Are you suggesting re-create it and delete at the test method level,
rather than at the test suite level ?

rockstar:
Yeah, that's what I'm suggesting. A test should be in charge of setting up its own resources, and then cleaning them up when they're done. The test should leave no trace that it ran.

jpic:
On Sun, May 29, 2016 at 7:34 AM, Paul Hummer <notifications@github.com>
wrote:

> Yeah, that's what I'm suggesting. A test should be in charge of setting up
> its own resources, and then cleaning them up when they're done. The test
> should leave no trace that it ran.
>
> Yes but if we create and upload a new image for every test, won't the
tests be too slow to be useful ?

rockstar:
I'm not concerned with how long the integration tests take. I realize that you're currently using them as unit tests, but that's not what they are intended for. They are intended to be a full end-to-end tests to validate LXD as much as pyLXD. It shouldn't matter how long they take. The unit tests are the ones designed to be the iterative test suite.

jpic:
I don't understand what "using integration tests as unit tests" means. For me, these are two different things, as shown in #98 I'm concerned by both. I'm not interrested in contributing something that will make them harder to use for me and others without any added value: none of the container test is changing anything on this image, and there's no reason to create it and delete it on each and every container test.

PR #101, opened 2016-05-24
Verify client connection

rockstar:
The client init will raise an exception if the endpoint isn't accessible. This fixes one of the two issues I've discovered while trying to sort out what's going on with my local lxd install.


zulcss:
lgtm

jpic:
didnt have time to review / discuss
best practice is to have a module level parent class
also it would help to have the response object in the expection

PR #102, opened 2016-05-25
Allow specification of devices in a profile

zulcss:
Currently its not possible to  specify devices when creating a LXD profile. This is required so we can attach
disks, network interfaces, etc in nova-lxd,

### DIFF
--- a/pylxd/profile.py
+++ b/pylxd/profile.py
@@ -44,11 +44,12 @@ class Profile(mixin.Marshallable):
         return profiles
 
     @classmethod
-    def create(cls, client, name, config):
+    def create(cls, client, name, config, devices={}):
         """Create a profile."""
         client.api.profiles.post(json={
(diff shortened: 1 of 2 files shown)


pylxd/profile.py
```diff
@@ -44,11 +44,12 @@ def all(cls, client):
         return profiles
 
     @classmethod
-    def create(cls, client, name, config):
+    def create(cls, client, name, config, devices={}):
```
rockstar: I'd prefer it `devices` was `None` by default, and then `if config is not None:` for the logic for adding devices.

jpic: Are you sure this is thread safe ? Doesn't looks so to me.

jpic:
Is there a typo in the commit message ?

> Currently its not specify devices when creating a LXD profile.

PR #103, opened 2016-05-25
Add more exceptions to the API

rockstar:
We've always known that our error handling has needed work. The original API over-caught things, and the new API doesn't catch anything. The new api's inability to catch things results in some really random errors (like, when `wait=True`, not finding the `operation` key in the returned dict).

We will likely need some sort of assertion on _every_ request we make, that makes sure we have the proper return statuses. We'll also need to think about the extra information we include in these new exceptions (a core reason why I try and use built-ins when I can).

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -14,7 +14,7 @@
 
 import six
 
-from pylxd import mixin
+from pylxd import exceptions, mixin
 from pylxd.containerState import ContainerState
 from pylxd.operation import Operation
 
@@ -38,7 +38,8 @@ class Container(mixin.Waitable, mixin.Marshallable):
         response = client.api.containers[name].get()
 
         if response.status_code == 404:
-            raise NameError('No container named "{}"'.format(name))
+            raise exceptions.NotFound(
+                'No container named "{}"'.format(name))
         container = cls(_client=client, **response.json()['metadata'])
         return container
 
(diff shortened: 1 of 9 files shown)


jpic:
Nice rewrite of #98, is it possible to have the response object too ?

pylxd/container.py
```diff
@@ -65,7 +66,7 @@ def create(cls, client, config, wait=False):
         response = client.api.containers.post(json=config)
 
         if response.status_code != 202:
-            raise RuntimeError('Error creating instance')
+            raise exceptions.CreateFailed()
```
jpic: Perhaps it would be useful to have the reason too, it's in the response object.

pylxd/exceptions.py
```diff
@@ -1,2 +1,10 @@
 class ClientConnectionFailed(Exception):
     """An exception raised when the Client connection fails."""
+
+
+class NotFound(Exception):
+    """Generic get failure exception."""
+
+
+class CreateFailed(Exception):
+    """Generic create failure exception."""
```
jpic: Oh, no parent module-level exception ? I thought it was a best practice and helped in exception handling ...

pylxd/container.py
rockstar: I don't really want to use the whole response, though I agree we need more information. The response json's message param would seem appropriate as a good start. I'll add it.

zulcss:
lgtm

PR #104, opened 2016-05-25
Allow specification of devices in a profile

zulcss:
Currently it not possible to specify devices when creating a
LXD profile. Make config and devices
optional when creating an LXD profile. This is required
so we can attach disks, network interfaces, etc.


rockstar:
👍

PR #108, opened 2016-05-25
Fix travis run for coverage

rockstar:


.travis.yml
```diff
@@ -9,13 +9,14 @@ env:
   # - TOXENV=integration # requires a remote lxd setup
 
 install:
-- travis_retry pip install tox
+  - travis_retry pip install tox
+  - pip install codecov
```
jpic: I recommend to add codecov above, ie: 

 - travis_retry pip install tox codecov

travis_retry is important here for the build not to fail in case of a network error - network is #1 enemy of CI ^^

:guitar:

Issue #79, opened 2016-05-16
Documentation lacking for accessing remote hosts


candlerb:
pylxd documentation doesn't say how to access remote hosts. Poking around, I find a number of issues.

1. The [front page](https://github.com/lxc/pylxd) intro document says to use `pylxd.api.API()`, however in the code pylxd.api is imported from pylxd.deprecated. Should anyone even be using this in the first place??

    Looking at [ansible lxd_container](https://github.com/ansible/ansible-modules-extras/pull/2208), they use `pylxd.client.Client()` instead. I haven't found any documentation for that entry point into pylxd. Indeed, if there is any documentation for pylxd, I've been unable to find it: only a skeleton at <https://github.com/lxc/pylxd/tree/master/doc/source>

2. From the source, `pylxd.api.API()` takes an optional host and port. I found that I can use `pylxd.api.API(host="nuc1.ws.nsrc.org").container_list()` and this works just fine. It doesn't matter which variation of the host's name I use.

3. From the source, `pylxd.client.Client()` takes an optional URL. However if I try this code:

    ~~~
    pylxd.client.Client('https://nuc1w.ws.nsrc.org:8443').containers.all()
    ~~~

    then I get:

    ~~~
    requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
    ~~~

    I have tried all the variations of the host name, and also connecting by IP address. If I connect using `openssl s_client` then I see the certificate has:

    ~~~
            Subject: O=linuxcontainers.org, CN=root@nuc1.ws.nsrc.org
    ...
                X509v3 Subject Alternative Name:
                    DNS:nuc1.ws.nsrc.org, DNS:fe80::c23f:d5ff:fe63:6411/64, DNS:10.10.0.238/24, DNS:10.10.0.237/32, DNS:2001:db8:100::238/64, DNS:fe80::c23f:d5ff:fe63:6411/64, DNS:fe80::fcd5:94ff:fe50:96ca/64, DNS:fe80::8461:ddff:fe6f:3aa8/64, DNS:fe80::1/64
    ~~~

    i.e. all hostname variations are in SAN (except I don't know why the certificate would include the prefix length, but then I've never seen certificates with IP addresses in them)

So anyway: for code which uses pylxd.client.Client() then I'd like to see an example of how to make it work without certificate errors.  I'd also like it to be clear whether pylxd.api or pylxd.client is the recommended interface to use.

[...]

PR #111, opened 2016-05-28
Add proper support for remote hosts

rockstar:
This branch does a number of things, all surrounding support for using LXD over HTTPS.

The first step was to handle self signed certs in LXD. By default, `python-requests` verifies server certificates. In most cases, this probably isn't going to work for LXD, because LXD generates its own self-signed certificates. `Client` now has a `verify` parameter, with a `True` default (because safety).

The second step was to add support for LXD's certificate authentication method. This meant adding a `cert` parameter to `Client`, which is a two part tuple of `(cert, key)`.

The last step was to document it all, essentially fixing #79.


zulcss:
LGTM

doc/source/usage.rst
```diff
@@ -10,18 +10,29 @@ localhost:
...
+
 This :class:`~client.Client` object exposes managers for:
 
 - :class:`~container.Container`,
 - :class:`~profile.Profile`,
 - :class:`~operation.Operation`,
 - :class:`~image.Image`,
 
-Also, it exposes the HTTP API with the `api <api.html#Client.api>`_ attribute,
-allowing lower-level operations.
```
jpic: There are a number of operations which we can't do with Client.

rockstar: Hi @jpic - the operations we can't do with Client (or the other APIs) should be implemented. The `api` property's existence is a byproduct of the architecture, but I don't want to encourage its use.

PR #112, opened 2016-05-29
Partial objects and updating

rockstar:
We currently have a bug where, if an object is fetched via `all`, it's not yet a complete object. We only have the single identifying property. It must first be populated before we can really interact with it. `Container` already had some support for this, but if one tried to call `update` on a partial object before fetching, it would result in a less than friendly `AttributeError` when trying to marshall the object. `Profile` and `Image` had to grow support for `fetch`, while `Container`'s api needed to be renamed (with the old name kept around for backwards compatibility.

### DIFF
--- a/pylxd/exceptions.py
+++ b/pylxd/exceptions.py
@@ -27,3 +27,15 @@ class NotFound(_LXDAPIException):
 
 class CreateFailed(_LXDAPIException):
     """Generic create failure exception."""
+
+
+class ObjectIncomplete(Exception):
+    """An exception raised when an object isn't completely populated.
+
+    When an object is fetched via `all`, it isn't a complete object
+    just yet. It requires a call to `fetch` to populate the object before
+    it can be pushed back up to the server.
+    """
+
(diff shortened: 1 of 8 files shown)


pylxd/exceptions.py
```diff
@@ -27,3 +27,15 @@ class NotFound(_LXDAPIException):
...
+class ObjectIncomplete(Exception):
+    """An exception raised when an object isn't completely populated.
+
+    When an object is fetched via `all`, it isn't a complete object
+    just yet. It requires a call to `fetch` to populate the object before
+    it can be pushed back up to the server.
+    """
+
+    def __str__(self):
+        return 'Object incomplete. Please call `fetch` before updating.'
```
jpic: Way cool error handling !!!

Do you want to add a test case to trigger a raise of this exception ?

jpic:
LGTM

zulcss:
lgtm

pjones-d:
Thanks!

PR #114, opened 2016-05-29
Add more documentation for pylxd

rockstar:
I was bored, so I expanded on the documentation. If there's a reason to add to the line count of this project, it's most certainly documentation.


rockstar:
I think I need to merge #113 into master and expand on those docs before I merge this patch, though a review would still be nice.

zulcss:
lgtm

PR #115, opened 2016-05-30
Snapshot manager

rockstar:
This patch adds a new api to `Container` that is similar to the `Client` `container`, `image`, and `profile` manager api. It also adds the use of these new apis to the old snapshot methods on `Container`, which should no longer be used.


rockstar:
I think I need to merge #114 into master and expand on those docs before I merge this patch, though a review would still be nice.

zulcss:
lgtm

PR #116, opened 2016-05-30
Deprecation warning

rockstar:
With 2.0, we committed to maintaining backwards compatibility for some amount of API stability. This is becoming more and more important as we expand our userbase. This patch adds a `deprecated` decorator, which emits a `DeprecationWarning` on any decorated object.

It's then used on `Container.reload`, which was replaced prior to 2.0.1, but not officially deprecated. All calls to `reload` have been replaced with `fetch`.

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -15,6 +15,7 @@
 import six
 
 from pylxd import exceptions, mixin
+from pylxd.deprecation import deprecated
 from pylxd.operation import Operation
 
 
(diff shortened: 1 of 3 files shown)


rockstar:
The codecov failure is a red herring. One call to `reload` was not covered previously, and so is not covered in this patch. Fixing it is beyond the scope of this patch, so while overall test coverage is increased, the patch's code coverage doesn't meet the current (very high) standard for code coverage.

I knew this would happen at some point. I'm happy that it happened sooner than expected.

zulcss:
lgtm

PR #119, opened 2016-05-30
Refactor the manager protocol to be a little cleaner

rockstar:
The best thing about a good API is that it's guessable, and that the process of extending it has a clear path. The manager methods we provided were nice, but it quickly became clear in #115 that we needed a better/quicker/cleaner way of adding a manager. This simplifies the process greatly.


rockstar:
Weird. CodeCov says everything is green, but also that code coverage will go down. It _appears_ that this is related to the fact that we'll have less lines of code. Hilarious, but totally fine.

zulcss:
lgtm

PR #120, opened 2016-05-31
Update the mock service to be more strict

rockstar:
I was writing some tests and discovered that, well, our mock service was allowing a little too much through. This should make it stricter by removing some wildcard regexes.


zulcss:
lgtm

PR #121, opened 2016-05-31
Files manager

rockstar:
This branch adds a manager-esque `FilesManager` to handle files, with accompanying tests.

Please review #120 first (I merged that into this branch; writing these tests were how I realized how loose the mock service was)

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -32,6 +32,31 @@ class Container(mixin.Waitable, mixin.Marshallable):
     This class is not intended to be used directly, but rather to be used
     via `Client.containers.create`.
     """
+    class FilesManager(object):
+        """A pseudo-manager for namespacing file operations."""
+
+        def __init__(self, client, container):
+            self._client = client
+            self._container = container
+
+        def put(self, filepath, data):
+            response = self._client.api.containers[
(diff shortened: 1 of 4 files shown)


pylxd/container.py
```diff
@@ -30,6 +30,31 @@ class Container(mixin.Waitable, mixin.Marshallable):
...
+                params={'path': filepath}, data=data)
+            return response.status_code == 200
+
+        def get(self, filepath):
+            response = self._client.api.containers[
+                self._container.name].files.get(
+                params={'path': filepath})
+            if response.status_code == 500:
+                # XXX: rockstar (15 Feb 2016) - This should really
+                # return a 404. I blame LXD.
```
stgraber: Can you file a bug against LXD when you see that kind of problem?

We indeed should be returning a 404 in such case, so it's probably best for your code to expect a 500 or 404 and have us fix the logic in LXD so it returns the right error code.

rockstar: Huh. I thought I had filed an issue (I remember talking about this in Cape Town). Anyway, see https://github.com/lxc/lxd/issues/2059

zulcss:
lgtm

PR #122, opened 2016-05-31
Add coverage exclusions

rockstar:
I added a commit hook that removed some cache files, like the coverage cache, and suddenly, I found a bunch of coverage stuff I had been missing. This patch adds some exclusions, like ignoring `__str__` methods. Also, apparently if you specify `exclude_lines`, you have to add the default `pragma: no cover` to it.


zulcss:
lgtm

PR #124, opened 2016-05-31
Cache LXD host info

zulcss:
Cache LXD host info so that it can be queried via the API.

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -131,7 +131,7 @@ class Client(object):
             if auth != "trusted":
                 raise exceptions.ClientAuthenticationFailed()
 
-            self.host_info = response.json()['metadata']['environment']
+            self.host_info = response.json()['metadata']
         except (requests.exceptions.ConnectionError,
                 requests.exceptions.InvalidURL):
             raise exceptions.ClientConnectionFailed()
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -130,6 +130,8 @@ class Client(object):
             auth = response.json()['metadata']['auth']
             if auth != "trusted":
                 raise exceptions.ClientAuthenticationFailed()
+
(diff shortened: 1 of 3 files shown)


pylxd/client.py
```diff
@@ -130,6 +130,8 @@ def __init__(self, endpoint=None, version='1.0', cert=None, verify=True):
             auth = response.json()['metadata']['auth']
             if auth != "trusted":
                 raise exceptions.ClientAuthenticationFailed()
+
+            self.host_info = response.json()['metadata']['environment']
```
rockstar: I would cache the entire `metadata` info itself. There are all sorts of things in that dict that are good for us to know for various reasons.

pylxd/tests/test_client.py
```diff
@@ -67,11 +70,18 @@ def raise_exception():
     def test_authentication_failed(self):
         """If the authentication fails, an exception is raised."""
         response = mock.MagicMock(status_code=200)
-        response.json.return_value = {'metadata': {'auth': 'untrusted'}}
+        response.json.return_value = {'metadata': {'auth': 'untrusted',
+                                                   'environment': {}
```
rockstar: You shouldn't need this mock here, as the exception should be raised before we get the metadata.

rockstar:
👍  With some small changes. Merge when those are complete.

pylxd/client.py
```diff
@@ -130,6 +130,8 @@ def __init__(self, endpoint=None, version='1.0', cert=None, verify=True):
             auth = response.json()['metadata']['auth']
             if auth != "trusted":
                 raise exceptions.ClientAuthenticationFailed()
+
```
rockstar: I'd remove the extra line here. No biggie one way or the other though, if you feel its important.

PR #125, opened 2016-06-01
Fix integration tests

rockstar:
This branch fixes the integration tests. There are a few things to address here:

  - The tests have no external dependencies, and require no outside setup.
  - The tests clean up after themselves (using `addCleanup`, with the exception of when something goes very wrong; this seems more a bug in `unittest` than our issue)
  - The integration tests take 20 seconds on my machine (addressing somewhat unfounded concerns about them taking "too long")
  - The integration tests uncovered two bugs: (1) The deprecated `Container.delete_snapshot` wasn't forwarding the `wait` parameter, and (2) `Profile.create` returns a 200, not a 202 (because it's a synchronous operation).

After this is merged, I'll connect with Stephane and get them connected to the lxd jenkins.

### DIFF
--- a/integration/test_containers.py
+++ b/integration/test_containers.py
@@ -11,11 +11,13 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+from pylxd import exceptions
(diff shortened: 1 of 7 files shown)


zulcss:
+1

PR #129, opened 2016-06-03
Better http assertions

rockstar:
This is a first pass at unifying the error handling of pylxd. The LXD API response codification is pretty tight, and it's pretty easy to figure out when something isn't right. I think there are probably a few more patches that can happen to help organize this a bit, but this is a good start.

Most of the patch is updating the tests to act more like the LXD API.


zulcss:
+1

Issue #127, opened 2016-06-02
No way to export images


rockstar:
We can't currently export images stored in lxd.

rockstar:
Fixed in #130

PR #130, opened 2016-06-03
Image export

rockstar:
Please review #129 first (this branch is based on that one; only the last two commits are relevant here)

This branch adds image export support, fixing issue #127


zulcss:
+1

PR #131, opened 2016-06-04
Add more tests for container state

rockstar:
While exploring the container state issue (that ended up being a non-issue), I added a bunch of test coverage...kinda. I mean, it executes the code path, but makes no assertions, which makes me slightly uncomfortable, but after pouring through the mock_services docs, it seemed non-trivial to handle that, and probably needs some re-working beyond the scope of a small branch.


zulcss:
+1

PR #132, opened 2016-06-05
Events websocket

rockstar:
Add support for the events websocket.

This was a little...special...in that we've mostly had this consistent API for working with a HTTP stateless api, and now we have what is essentially a socket client. I tried to document this as clearly as possible, but it's entirely possible that we'll have to iterate on this API a few times before it "feels right".


zulcss:
+1

PR #133, opened 2016-06-06
Add support for container migration

zulcss:
Add support for container migration

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -171,6 +171,12 @@ class Container(mixin.Waitable, mixin.Marshallable):
         state = ContainerState(**response.json()['metadata'])
         return state
 
+    def migrate(self):
+        response = self._client.api.containers[self.name].post(json={
+            'migration': True})
+        migrate = ContainerState(**response.json()['metadata'])
+        return migrate
+
     def start(self, timeout=30, force=True, wait=False):
         """Start the container."""
         return self._set_state('start',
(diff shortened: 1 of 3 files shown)


pylxd/tests/test_container.py
```diff
@@ -141,6 +141,13 @@ def test_delete(self):
 
         an_container.delete(wait=True)
 
+    def test_migrate(self):
+        """A container is migrate."""
+        name = 'an-container'
+
+        an_container = container.Container.get(self.client, name)
+        an_container.migrate()
```
rockstar: This isn't a good test, as it doesn't return assert anything (I realize that we have _some_ tests that don't assert things, but we have a mock service that has some support, or the test would fail).

pylxd/tests/test_container.py
```diff
@@ -141,6 +141,13 @@ def test_delete(self):
 
         an_container.delete(wait=True)
 
+    def test_migrate(self):
+        """A container is migrate."""
```
rockstar: "A container is migrated."

pylxd/container.py
```diff
@@ -171,6 +171,12 @@ def state(self):
         state = ContainerState(**response.json()['metadata'])
         return state
 
+    def migrate(self):
```
rockstar: You probably need a docstring here.

rockstar:
This patch doesn't actually add support for migration. It has some more work needed before it supports migration.

PR #134, opened 2016-06-07
Execute interactive

rockstar:
This patch changes the semantics of `Container.execute` a bit. Now it can _only_ take a list as a command argument (this is required as a lxd implementation detail). It also returns the stdout and stderr as text from the command (i.e. it now blocks waiting for the command to be run). This required some...adventurous...websocket work. If anything doesn't seem clear, just let me know and I'll try and add some documentation.


zulcss:
lgtm

PR #135, opened 2016-06-07
Operation work

rockstar:
Simplify the `Operation` handling work.

In general, operational work is entirely an implementation detail that we want to hide from the user. However, I made these two changes because it just made it easier for me as a developer (and reduced some repeated code).


zulcss:
lgtm

PR #136, opened 2016-06-09
Client certificates

rockstar:
This branch adds support for authenticating a non-trusted 'Client`. Previously, the client would raise an exception when the client is untrusted. Now, it's possible to create a `Client` and authenticate it after the fact.

Under the covers, it uses the same "manager" pattern that other objects have used, though it's probably less likely that most users will ever need to use `client.certificates.all`, etc.


tych0:
LGTM!

PR #139, opened 2016-06-12
Fix ValueError when HTTP_PROXY env var is present

### LINKED ISSUE #138: pylxd.Client() raises ValueError if e.g. http_proxy environment variable is set
pylxd appears to pass any proxy related environment variables onto Requests, even in the default case when connecting to a Unix domain socket. This results in an an unhandled exception
```
$ python
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pylxd
>>> pylxd.__version__
'2.0.2'
>>> pylxd.Client()
  File "/home/willmerae/.local/lib/python2.7/site-packages/pylxd/client.py", line 183, in __init__
    response = self.api.get()
  File "/home/willmerae/.local/lib/python2.7/site-packages/pylxd/client.py", line 87, in get
    response = self.session.get(self._api_endpoint, *args, **kwargs)
  File "/home/willmerae/.local/lib/python2.7/site-packages/requests/sessions.py", line 487, in get
    return self.request('GET', url, **kwargs)
  File "/home/willmerae/.local/lib/python2.7/site-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/willmerae/.local/lib/python2.7/site-packages/requests/sessions.py", line 585, in send
    r = adapter.send(request, **kwargs)
  File "/home/willmerae/.local/lib/python2.7/site-packages/requests/adapters.py", line 370, in send
    conn = self.get_connection(request.url, proxies)
  File "/home/willmerae/.local/lib/python2.7/site-packages/requests_unixsocket/adapters.py", line 53, in get_connection
    % self.__class__.__name__)
ValueError: UnixAdapter does not support specifying proxies
```

If I unset every proxy related environment variable, then the call succeeds

```
$ unset http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY 
$ python
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pylxd
>>> pylxd.Client()
<pylxd.client.Client object at 0x7f6d65fd5e10>
```

moreati:
Fixes #138


setup.py
```diff
@@ -28,6 +28,6 @@
     setup_requires=[
         'pbr>=1.8',
         'requests!=2.8.0,>=2.5.2',
-        'requests-unixsocket==0.1.4',
+        'requests-unixsocket>=0.1.5',
```
rockstar: Could you add a comment as to why we need `0.1.5`? That would make it clear that there's a hard dependency here (whereas, `0.1.4` was there because it was the released version when we first started using `request-unixsocket`. Maybe something like `# Needed for HTTP_PROXY support`

tox.ini
```diff
@@ -17,7 +17,11 @@ commands = nosetests --with-coverage --cover-package=pylxd pylxd
 commands = flake8 --ignore=E123,E125
 
 [testenv:integration]
-commands = nosetests integration
+whitelist_externals =
+    env
+commands =
+    env HTTP_PROXY='http://127.0.0.42' nosetests integration.test_proxies
```
rockstar: I appreciate that this validates the fix, but I don't think it's needed. If the fix is in a dependency, we don't need to test the dependency. I don't think this part of the patch is needed.

rockstar:
Thank you for the patch! I hadn't had time to chase it, and it's always helpful to have more contributors. I had two small comments for you, but I'll merge the patch as soon as those comments are addressed. Again, thanks for chasing this.

rockstar:
Thank you so much for the patch!

PR #144, opened 2016-06-25
Stateful object experiment

rockstar:
We've talked a few times about how LXD object representations would be able to keep track of their state. For instance, has the instance been modified since it was retrieved from the server? If not, there's no reason to call `update`.

I also tried to use this time to sort out the object lifecycle as well. There's the creation (done via `get` \ `create`), what is currently referred to as `fetch` but is being reconsidered as `sync` (obviously, with a `DeprecationWarning` when implemented in the other objects) to sync up a partial object gotten via `all`, `update` (being renamed to `save`) for putting changes back into the server, and `delete` for...deleting.

I wanted to take this experiment into the realm of practical, so I implemented `/1.0/networks`, which was the last endpoint that hasn't been implemented yet. Unfortunately, there wasn't much to networks (no create, update, or delete), so it didn't end up being super practical.

This is pretty complicated. It is. I'll admit it. I think it's a little justified for a few reasons: one, it allows us to catch errors a little earlier and some flexibility. The `Attribute.validator` property, for instance, could check types, we could add more support for read only attributes (like `name` in many cases), or we could add better support for marshalling/de-marshalling (for instance, methods that convert `Network.used_by` to list of `Container`s and back to a list of strings). It also gives us a mechanism for handling the Etag support that just landed in LXD.

Comments? Suggestions? Curses?

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -194,6 +194,7 @@ class Client(object):
         self.certificates = managers.CertificateManager(self)
         self.containers = managers.ContainerManager(self)
         self.images = managers.ImageManager(self)
+        self.networks = managers.NetworkManager(self)
         self.operations = managers.OperationManager(self)
         self.profiles = managers.ProfileManager(self)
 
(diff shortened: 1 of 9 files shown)


zulcss:
lgtm

PR #151, opened 2016-06-28
Attribute state

rockstar:
This patch adds the ability for the model to know what changed since the last sync. This would allow us to do things like use PATCH instead of PUT. It also added some transactional elements, like the ability to rollback all your local changes to whatever is on the server.

I also found a bug in the marshalling around `Attribute.readonly`. That is now fixed.


zulcss:
+1

PR #153, opened 2016-06-28
Extraneous exceptions

rockstar:
- Remove the extra exceptions that aren't really needed with the introduction of `LXDAPIException`.
- Fix the assertions of responses for POST, to support LXD 2.0.3


zulcss:
+1

PR #157, opened 2016-07-13
Add support for container and snapshot publish.

rockstar:
Snapshot publish is broken in LXD, but this is a guess for how
it is supposed to work, based on how it works in container
publish.

### DIFF
--- a/pylxd/container.py
+++ b/pylxd/container.py
@@ -37,6 +37,7 @@ class Container(model.Model):
     This class is not intended to be used directly, but rather to be used
     via `Client.containers.create`.
     """
+
     architecture = model.Attribute()
     config = model.Attribute()
     created_at = model.Attribute()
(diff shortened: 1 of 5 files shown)


rockstar:
Ignore the codecov failures: we're executing every line, and I'm not concerned about partials.

pylxd/container.py
```diff
@@ -374,3 +398,28 @@ def rename(self, new_name, wait=False):
...
+        https://github.com/lxc/lxd/issues/2201 - The implementation
+        here is mostly a guess. Once that bug is fixed, we can verify
+        that this works, or file a bug to fix it appropriately.
+        """
+        data = {
+            'public': public,
+            'source': {
+                'type': 'snapshot',
+                'name': '{}/{}'.format(self.container.name, self.name),
+                'alias': 'lololol',
```
zulcss: Is this necessary?

zulcss:
lgtm

PR #161, opened 2016-07-24
Quote the path of unix endpoints.

pcdummy:
Allow to specify an existing socket as endpoint and quote it for requests.unixsocket.

### DIFF
--- a/pylxd/tests/test_client.py
+++ b/pylxd/tests/test_client.py
@@ -13,6 +13,7 @@
 #    under the License.
 import json
 import os
+import os.path
 import unittest
 
 import mock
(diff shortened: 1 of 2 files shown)


rockstar:
It looks like you need to add some tests for this functionality. `pylxd.tests.test_client` should have the framework for writing the tests.

pylxd/tests/test_client.py
```diff
@@ -72,6 +73,17 @@ def test_create_endpoint(self):
 
         self.assertEqual(expected, an_client.api._api_endpoint)
 
+    def test_create_endpoint_unixsocket(self):
+        """Test with unix socket endpoint."""
+        endpoint = '/tmp/unix.socket'
+        expected = 'http+unix://%2Ftmp%2Funix.socket/1.0'
+
+        if not os.path.isfile(endpoint):
+            open(endpoint, 'a').close()
```
pcdummy: I'm not very happy about this "solution" what do you think  @rockstar

rockstar: You can mock out `os.path.isfile` to return True in this case. That'd be preferred. We don't want to have to touch the filesystem in tests at all if we can help it.

pcdummy: Done, thank you!

PR #163, opened 2016-07-24
Return self on container state changes.

pcdummy:
This restores the previous behaviour.


rockstar:
What "previous behavior" are you referring to? I'm unsure what the need for this patch is, and I would like to understand. I don't particularly think that we should be returning an http response object ever, and I'm not sure why we would return `self` in this case either.

pcdummy:
I was wrong, thought previously state changes returned the updated container, they don't. Sorry.

PR #164, opened 2016-07-25
Add images.copy()

pcdummy:
This is a wip but @rockstar please review and give your thoughts.

### DIFF
--- a/pylxd/image.py
+++ b/pylxd/image.py
@@ -17,6 +17,22 @@ from pylxd import model
 from pylxd.operation import Operation
 
 
+def _image_create_from_config(client, config, wait=False):
+    """ Create an image from the given configuration.
+
+    See: https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-6
+    """
+    if 'source' not in config or 'fingerprint' not in config['source']:
+        raise KeyError(
+            'You need to provide a fingerprint '
+            'in the config[\'source\'] argument'
+        )
+
+    response = client.api.images.post(json=config)
+    if wait:
+        Operation.wait_for_operation(client, response.json()['operation'])
+
+
 class Image(model.Model):
     """A LXD Image."""
     aliases = model.Attribute(readonly=True)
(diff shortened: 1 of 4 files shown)


pcdummy:
@rockstar please have a look.

pylxd/image.py
```diff
@@ -80,7 +81,66 @@ def create(cls, client, image_data, public=False, wait=False):
             Operation.wait_for_operation(client, response.json()['operation'])
         return cls(client, fingerprint=fingerprint)
 
+    @classmethod
+    def create_from_config(cls, client, config, wait=False):
```
rockstar: I don't really care for the "pass in a random config as a parameter", particularly because later on, there are requirements enforced on the config itself. I think I'd rather have a `copy_from_host` and 'create_from_simplestream` class methods. For example, `copy_from_host` would take a `name` and another `client` as parameters, and already that makes for a cleaner interface.

pylxd/image.py
```diff
@@ -80,7 +81,66 @@ def create(cls, client, image_data, public=False, wait=False):
...
+        if wait:
+            Operation.wait_for_operation(client, response.json()['operation'])
+        return cls(client, fingerprint=config['source']['fingerprint'])
+
     def export(self):
         """Export the image."""
         response = self.api.export.get()
         return response.content
+
+    def copy_to_lxd(self, new_client, public=None, auto_update=None,
```
rockstar: I would certainly love this functionality (and the same with `Profile`). I think `copy` is a better name for this, but the function definition is probably fine. Is there a reason we can't get `public` and `auto_update` from the properties from the existing image, rather than specifying them here? If not, `public` should default to `False`, rather than `None`.

pylxd/image.py
```diff
@@ -80,7 +81,66 @@ def create(cls, client, image_data, public=False, wait=False):
...
+                    wait=False):
+        """ Copy an image to a another LXD.
+
+        Destination host information is contained in the client
+        connection passed in.
+        """
+        self.sync()  # Make sure the object isn't stale
+
+        # TODO: this is a hacky way to remove the /1.0 do we have a better one?
```
rockstar: There isn't a better way, unfortunately. I'd love to have one. You're already doing something slightly hacky by grabbing `_api_endpoint`, and I do something similar for websocket handling.

pylxd/image.py
```diff
@@ -80,7 +81,66 @@ def create(cls, client, image_data, public=False, wait=False):
...
+        url = '/'.join(self.client.api._api_endpoint.split('/')[:-1])
+
+        if public is None:
+            public = self.public
+
+        if auto_update is None:
+            auto_update = self.auto_update
+
+        config = {
```
rockstar: I like this pattern here. This is what I mean by not asking the user to create a random dict. We can construct this with our knowledge of LXD based on there parameters, and then the user doesn't need to know everything about LXD.

pylxd/image.py
```diff
@@ -80,7 +81,66 @@ def create(cls, client, image_data, public=False, wait=False):
...
+        Destination host information is contained in the client
+        connection passed in.
+        """
+        self.sync()  # Make sure the object isn't stale
+
+        # TODO: this is a hacky way to remove the /1.0 do we have a better one?
+        url = '/'.join(self.client.api._api_endpoint.split('/')[:-1])
+
+        if public is None:
+            public = self.public
```
rockstar: Oh, I see why `public` is `None` now.

pylxd/image.py
pcdummy: This method is for the `copy_to_lxd` method, `copy_to_lxd` generates the `config` for this and calls 
`newclient.images.create_from_config(`generated_config`, wait=`given wait`)

It would also allow someone to create images from url and simplestreams, but for that we want implicit methods i think.

rockstar: Ah, I see now. It's being called from the other method you're creating. A couple of things here:

- This doesn't need to be a method. It can be a function not contained scoped in `Image`.
- This shouldn't be a public function. Preface it with a `_`. We want to be able to modify it's usage without breaking people (and if people are using functions prefaced with `_`, I can't care about them).

pylxd/image.py
rockstar: Oh, I forgot to point out that we don't use `TODO` for acknowledging technical debt. You should use this format:

    # XXX: pcdummy (28 Jul 2016) - This is a hacky way....

The formatting is important (including the date), as it is consistent with the rest of the project and many tools.

PR #166, opened 2016-07-25
Add Attribute.optional, this fixes auto_update with images.linuxcontainers.org

### LINKED ISSUE #165: AttributeError: last_used_at on images.linuxcontainers.org
Doing the following:

```python
import pylxd
c1 = pylxd.Client(endpoint='https://images.linuxcontainers.org:8443', cert=('/root/.config/lxc/client.crt', '/root/.config/lxc/client.key',), verify=False)
c1.images.get_by_alias('ubuntu/xenial/amd64').marshall()
```

results in
```
  File "/usr/local/lib/python2.7/dist-packages/pylxd/model.py", line 183, in marshall
    marshalled[key] = getattr(self, key)
  File "/usr/local/lib/python2.7/dist-packages/pylxd/model.py", line 113, in __getattribute__
    return super(Model, self).__getattribute__(name)
AttributeError: last_used_at
```

pcdummy:

### DIFF
--- a/pylxd/image.py
+++ b/pylxd/image.py
@@ -20,18 +20,18 @@ from pylxd.operation import Operation
 class Image(model.Model):
     """A LXD Image."""
     aliases = model.Attribute(readonly=True)
-    auto_update = model.Attribute()
-    architecture = model.Attribute()
-    cached = model.Attribute()
-    created_at = model.Attribute()
-    expires_at = model.Attribute()
-    filename = model.Attribute()
-    fingerprint = model.Attribute()
-    last_used_at = model.Attribute()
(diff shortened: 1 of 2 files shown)


rockstar:
I think the way to fix this one is to set `last_used_at` as readonly.

pcdummy:
With these fields that might work, but what is it with "auto_update" which isn't readonly and doesn't get delivered by images.linuxcontainers.org ?

rockstar:
Hm, maybe we need to add an `optional` parameter, and if it's not set and is `optional`, we can successfully skip it. I'd like to opt in to the optional attributes, rather than just assume everything is optional.

PR #167, opened 2016-07-25
Add dict method which includes readonly attributes to the model.

pcdummy:
[See](https://github.com/pcdummy/saltstack-lxd-formula/commit/eed61b2617841fbf10400b3d499ef7145e25dffe) it in action

### DIFF
--- a/pylxd/model.py
+++ b/pylxd/model.py
@@ -182,3 +182,11 @@ class Model(object):
             if not val.readonly:
                 marshalled[key] = getattr(self, key)
         return marshalled
+
+    def dict(self):
+        """ Same as marshall but includes readonly attributes."""
+        marshalled = {}
+        for key, val in self.__attributes__.items():
+            if hasattr(self, key):
+                marshalled[key] = getattr(self, key)
+        return marshalled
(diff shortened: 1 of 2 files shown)


rockstar:
Can you explain why we would need this? I realize you're using it elsewhere in another library, but I can't think of why this logic should live in pylxd. In general, even the `marshall` method is kind of a "private" method, in that it's not really part of the public/documented api, and generally you shouldn't even need to call it yourself.

pcdummy:
For me this dict method is handy as i don't need to know which `__attributes__` are stored on a model and
its a easy method to convert any PyLXD object to a dict, which is nice for json/yaml output on salt.

I could write my own "object_to_dict" but then i need to inspect it or iterrate over obj.__attributes__ which i'm sure we both don't want.

rockstar:
While I appreciate your effort, I'm not sure that this is really suited for pylxd. A complication comes from the fact that you're using the objects in a little different way than pylxd itself is. I don't think iterating over `__attributes__` in your code is unrealistic (and we've defined a pretty good framework for that). This code, itself, would be unused by pylxd itself.

pcdummy:
So as a solution you encourage me to iterate over __attributes__, right?

PR #168, opened 2016-07-25
Add image aliases support.

pcdummy:
As discussed in the IRC with @rockstar here comes the IAS - Image Alias Support.

### DIFF
--- a/pylxd/image.py
+++ b/pylxd/image.py
@@ -19,7 +19,7 @@ from pylxd.operation import Operation
 
 class Image(model.Model):
     """A LXD Image."""
-    aliases = model.Attribute()
+    aliases = model.Attribute(readonly=True)
     auto_update = model.Attribute()
     architecture = model.Attribute()
     cached = model.Attribute()
@@ -84,3 +84,33 @@ class Image(model.Model):
         """Export the image."""
         response = self.api.export.get()
         return response.content
+
+    def add_alias(self, name, description):
+        """Add an alias to the image."""
+        self.client.api.images.aliases.post(json={
+            'description': description,
+            'target': self.fingerprint,
+            'name': name
+        })
+
+        # Update current aliases list
+        self.aliases.append({
+            'description': description,
+            'target': self.fingerprint,
+            'name': name
+        })
+
+        return self
+
+    def delete_alias(self, name):
+        """Delete an alias from the image."""
+        self.client.api.images.aliases[name].delete()
+
+        # Update current aliases list
+        la = [a['name'] for a in self.aliases]
+        try:
+            del self.aliases[la.index(name)]
+        except ValueError:
+            pass
+
+        return self
--- a/pylxd/image.py
+++ b/pylxd/image.py
@@ -100,8 +100,6 @@ class Image(model.Model):
             'name': name
         })
 
-        return self
-
     def delete_alias(self, name):
         """Delete an alias from the image."""
         self.client.api.images.aliases[name].delete()
@@ -112,5 +110,3 @@ class Image(model.Model):
             del self.aliases[la.index(name)]
         except ValueError:
             pass
-
-        return self
--- a/doc/source/images.rst
+++ b/doc/source/images.rst
@@ -49,6 +49,9 @@ Image methods
   - `export` - Export the image. Returns binary data that is the
     image itself.
 
+  - `add_alias` - Add an alias to the image.
+
+  - `delete_alias` - Remove an alias.
 
 Examples
 --------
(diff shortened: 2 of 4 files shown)


pylxd/image.py
```diff
@@ -84,3 +84,25 @@ def export(self):
...
+    def add_alias(self, name, description):
+        """Add an alias to the image."""
+        self.client.api.images.aliases.post(json={
+            'description': description,
+            'target': self.fingerprint,
+            'name': name
+        })
+
+        # Update current aliases list
+        self.sync()
```
rockstar: Rather than call `sync` here, you should just add the alias to the object locally. Doing a sync will wipe away any local changes (which is a known issue), but it seems an awkward side effect that adding an alias loses local/unsync'd changes.

pylxd/image.py
```diff
@@ -84,3 +84,25 @@ def export(self):
...
+
+        return self
+
+    def delete_alias(self, name):
+        """Delete an alias from the image."""
+        self.client.api.images.aliases[name].delete()
+
+        # Update current aliases list
+        self.sync()
```
rockstar: Same here with sync.

pcdummy: Want to only allow the deletion of aliases in the current image?

pylxd/image.py
```diff
@@ -84,3 +84,33 @@ def export(self):
...
+
+    def delete_alias(self, name):
+        """Delete an alias from the image."""
+        self.client.api.images.aliases[name].delete()
+
+        # Update current aliases list
+        la = [a['name'] for a in self.aliases]
+        try:
+            del self.aliases[la.index(name)]
+        except ValueError:
```
rockstar: You're missing code coverage here. I wonder why it is that we would need to add the `ValueError` here. When would we hit a case where the alias doesn't exist? If there are valid cases, we should list them in a comment here.

pylxd/image.py
```diff
@@ -84,3 +84,33 @@ def export(self):
...
+        })
+
+        # Update current aliases list
+        self.aliases.append({
+            'description': description,
+            'target': self.fingerprint,
+            'name': name
+        })
+
+        return self
```
rockstar: Per the other review, why would we need to return self here? It's an instance method, so I'm unsure what the reasoning was here.

pylxd/image.py
```diff
@@ -84,3 +84,33 @@ def export(self):
...
+        self.client.api.images.aliases[name].delete()
+
+        # Update current aliases list
+        la = [a['name'] for a in self.aliases]
+        try:
+            del self.aliases[la.index(name)]
+        except ValueError:
+            pass
+
+        return self
```
rockstar: Same here. Why return `self`?

pylxd/image.py
pcdummy: Currently `delete_alias` allows you delete any aliases, so we need to check after if we have this Alias for the current image.

pylxd/image.py
pcdummy: `self` allows you to:
```python
image.delete_alias("a").delete_alias("b")
```

This isn't a feature i realy need.

rockstar: If you don't really need the feature, the issue is moot, and we could just delete the line.

It would seem awkward (and maybe non-pythonic) to do `image.delete_alias("a").delete_alias("b")`. You aren't making any better code than something like:

    image.delete_alias('a')
    image.delete_alias('b')

...or better yet...

    for alias in ['a', 'b']:
        image.delete_alias(alias)

pylxd/image.py
rockstar: Have you tested this? I would expect that `self.client.api.images.aliases[name].delete()` would raise an exception if the alias doesn't exist. It's possible that you get a safer exception that way anyway, and that this exception handling is probably okay.

pcdummy: Think we speak about different kind of things.

`self.client.api.images.aliases[name].delete()` deletes any aliases if they are on the image or not,
so if theres an alias "an-alias" on image "a" but we delete it on image "b" the delete() function would success but the index method would throw a ValueError.

rockstar: I _think_ that if you try and delete LXD image aliases that don't exist, `LXDAPIException` will be raised with a 404. That's what I'm trying to say. I'd suggest trying this against a real LXD instance and see what happens.

pcdummy: I did so and yes `LXDAPIException` will be raised. BUT as said you can delete *any alias* from *any image*, means the alias *maybe not in the current image* this is why i have to catch ValueError.

pcdummy:
@rockstar please let me squash this before merge.

pylxd/image.py
rockstar: Yeah, that totally makes sense. Just needed test coverage for it, and it looks like you have that now.

PR #170, opened 2016-08-12
Adding feature of update container configuration

rooty0:


rooty0:
I just found method test_update which actually should do the same job as code in this commit. 
So the question is this ever been implemented?

rooty0:
nm, found it, undocumented :(

PR #174, opened 2016-09-07
Warn but don't raise exception on unknown attributes.

rockstar:
LXD adding new properties shouldn't cause current applications to fail,
but it is appropriate for pylxd to emit warnings about them.


zulcss:
Please merge once passes.

PR #175, opened 2016-09-07
Move models

rockstar:
Move the models to `pylxd.models` - This way, we separate what our data model is from what our API is


zulcss:
+1

PR #176, opened 2016-09-07
Fix integration tests

rockstar:
The integration tests were broken. While fixing them, I also changed `Model.__dirty__` to be a set, not a list.


zulcss:
+1

PR #178, opened 2016-09-13
Add script to easily run the integration tests.

rockstar:
The integration tests require a pretty clean environment, so running
them in my dev environment gets...weird sometimes. This will fire up
a container and run it inside that container. The caveat is that the
container is still (currently) privileged, so don't just run anyone's
integration tests on your machine. Look at them first.


zulcss:
+1

PR #179, opened 2016-09-16
Don't allow local connection clients to migrate

rockstar:
If a connection to LXD is a local connection, it should not be used to migrate, because it won't have a routable connection.

Originally, I had the idea that we could iterate through all the addresses in `metadata.environment.addresses` until we found a valid one, and while that made me laugh at the stupidity and possible genius of the hack, I think it's better to explicitly put the burden on the user to have a routable connection.


zulcss:
lgtm

PR #188, opened 2016-10-17
Add missing attributes to Network model

hsoft:
Add "config" and "managed" attributes to the Network model. In addition
to be generally useful information, it also fixes warnings we were
getting upon calling `Client.networks.get()` about these attributes not
existing.

### DIFF
--- a/pylxd/models/network.py
+++ b/pylxd/models/network.py
@@ -19,6 +19,8 @@ class Network(model.Model):
     name = model.Attribute()
     type = model.Attribute()
     used_by = model.Attribute()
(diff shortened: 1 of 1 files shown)


stgraber:
Just an outsider's observation, but why is pylxd warning/erroring on missing attributes to begin with?

We make a promise of API backward compatibility in LXD but we can absolutely add new endpoints and attributes without breaking backward compatibility and we in fact do so pretty regularly. It'd be nice if we could keep doing that without causing trouble for pylxd users.

PR #192, opened 2016-10-21
Extract the migration data step of migration.

rockstar:
This means that this data can be re-used in cases where the client
is a unix socket. The API consumer can then massage the data
accordingly.


zulcss:
lgtm

PR #193, opened 2016-10-25
Fix the ws4py connection handling for container execution

### LINKED ISSUE #182: Call to execute() is slow
Using pylxd 2.1, it seems the execute() call is quite slow and takes at least 5 seconds to finish. Not sure if its due to ws4py or something else.

rockstar:
This branch does a number of things. First, it simplifies _StdinWebsocket,
as we don't need most of what existed there. It closes immediately
after connection (and the command doesn't execute unless stdin is
also connected).

_CommandWebsocketClient learned how to pre-emptively close a connection.
We were previously waiting for the command to time out, which is why
the execute call was taking so long. This fixes #182.

The while loop that waits for the commands to complete had too long of
a delay, and had too much logic for whether the loop should continue
to wait. That while loop was also simplified.

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -313,9 +313,6 @@ class _CommandWebsocketClient(WebSocketBaseClient):  # pragma: no cover
         else:
             self.buffer.append(message.data.decode('utf-8'))
 
-    def closed(self, code, reason=None):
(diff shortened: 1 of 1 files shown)


rockstar:
Ignore the codecov failure. I can't realistically test the websocket stuff anyway.

zulcss:
lgtm

PR #194, opened 2016-10-25
Fully deprecate the old api

rockstar:
At this point, we should deprecate the code so far as to see that
anywhere it's being used has "pylxd.deprecated" in the import. I
think this makes sense for nova-lxd, but also for any code still
using the old API.


zulcss:
lgtm

PR #198, opened 2016-11-16
Add Container.execute_with_result

rockstar:
Container.execute needs a sane, safe way to return more properties
than just stdout/stderr. Container.execute_with_result will return
a ContainerExecuteResult with the exit code, stdout and stdin, and
provides a mechanism for returning more properties in the future.

Container.execute now raises a DeprecationWarning as its return
value will change in 2.2


zulcss:
lgtm

PR #200, opened 2016-11-17
Authenticate fixes

rockstar:
A number of bugs all conspired to make sure authentication didn't work (and I'm not sure how it ever worked...) This branch fixes them all and adds an integration test to make sure it doesn't break again.

  - `pylxd.Client` had no `cert` attribute, which broke `authenticate`
  - If the endpoint url specified in `pylxd.Client` creation ended in a trailing slash, LXD was responding with 301 redirects (appropriately) which also broke authentication
  - The integration tests needed to be modified to support listening on the public http interface to test the authentication mechanism.

### DIFF
--- a/integration/test_client.py
+++ b/integration/test_client.py
@@ -1,0 +1,35 @@
+# Copyright (c) 2016 Canonical Ltd
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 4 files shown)


rockstar:
Ignore the code coverage issue. It's a set of functionality that can only really be tested via integration test, which this patch provides.

zulcss:
lgtm

Issue #172, opened 2016-08-16
Image upload fails with multi part file (http upload)


zulcss:
```
pylxd.exceptions.ClientConnectionFailed
>>> c = Client()
>>> image_data = open('lxd_post_220970285').read()
>>> i = c.images.create(image_data, wait=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/stack/pylxd/pylxd/image.py", line 94, in create
    Operation.wait_for_operation(client, response.json()['operation'])
  File "/opt/stack/pylxd/pylxd/operation.py", line 29, in wait_for_operation
    operation.wait()
  File "/opt/stack/pylxd/pylxd/operation.py", line 51, in wait
    raise exceptions.LXDAPIException(response)
pylxd.exceptions.LXDAPIException: detectCompression failed, err='Unsupported compression.', tarfile='/var/lib/lxd/images/lxd_build_064453356/lxd_tar_482497790'
```

rockstar:
Can you import this image using the `lxc` command? I suspect this API is DTRT.

rockstar:
What format is this image in? Where did you get it? Can you put it somewhere I can get to it, so I can try it out?

rockstar:
I have been able to reproduce this issue, but when I do `lxc image import <file>`, I get `error: detectCompression failed, err='Unsupported compression', tarfile='/var/lib/lxd/images/lxd_build_626862389/lxd_tar_962061103'`

This looks like it's acting as it should. If you have an image that causes the issue, but successfully imports to LXD, please feel free to link it here and re-open.

rockstar:
Oh, I figured out what the issue is here. It's the combination of `metadata.yaml` and the image in the same request. That's something pylxd doesn't currently do, but should. I'll have to think about what the api for that looks like.

PR #203, opened 2016-11-23
Allow multipart form uploads for images.

rockstar:
In some cases, a user may want to upload a manifest with a raw image,
and pylxd previously didn't support that. This patch adds the needed
functionality.

As I did this, I found a bug in Image.create where fingerprint isn't
always reliably determined, so I have deprecated the ability to pass
a wait parameter to Image.create.

Fixes issue #172


zulcss:
lgtm

PR #204, opened 2016-11-23
Support more cryptography versions.

rockstar:
A request came in to audit pylxd requirements to line up better
with Openstack. There isn't much overhead to maintaining it with
cryptography, as those developers are pretty responsible.


zulcss:
lgtm

PR #205, opened 2016-11-24
Add `NotFound` and `exists`

rockstar:
A common API issue that users report is an easier way to detect when an object just doesn't exist. This is becoming more and more apparent as I update methods in nova-lxd.

Add `NotFound` in the case of 404, which inherits from `LXDAPIException`, so it doesn't break backwards compatibility.

`Container`, `Image`, and `Profile` now have a `exists` class method, so that one can do `client.containers.exists('an-container')` as a wrapper around this new functionality.

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -64,6 +64,8 @@ class _APINode(object):
         user.
         """
         if response.status_code not in allowed_status_codes:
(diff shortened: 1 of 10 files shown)


zulcss:
lgtm

PR #212, opened 2016-12-17
Check for unix socket to set missing scheme

keesbos:
The os.path.isfile(path) returns False if path is a unix socket

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -201,6 +201,9 @@ class Client(object):
             ):
                 self.api = _APINode('http+unix://{}'.format(
                     parse.quote(endpoint, safe='')))
+            elif endpoint.startswith('http+unix://') and '/' in endpoint[12:]:
+                self.api = _APINode('http+unix://{}'.format(
+                    parse.quote(endpoint[12:], safe='')))
             else:
                 # Extra trailing slashes cause LXD to 301
                 endpoint = endpoint.rstrip('/')
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -14,6 +14,7 @@
 import json
 import os
 import os.path
+import stat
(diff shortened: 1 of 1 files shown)


keesbos:
BTW. I don't understand why os.path.isfile should be used anyway (returns only True for regular files and if it's a symbolic link to a regular file).

In other words, I think only unix socket files are/should be supported and the os.path.isfile is to be removed.

pylxd/client.py
```diff
@@ -190,9 +191,19 @@ class Client(object):
...
+                endpoint.startswith('/') and (
+                    os.path.isfile(endpoint) or (
+                        os.path.exists(endpoint) and
+                        stat.S_ISSOCK(os.stat(endpoint).st_mode)
+                    )
+                )
+            ):
                 self.api = _APINode('http+unix://{}'.format(
                     parse.quote(endpoint, safe='')))
+            elif endpoint.startswith('http+unix://') and '/' in endpoint[12:]:
```
rockstar: I would rather not expose the `http+unix://` part of the url to api consumers. Recently, I've had my faith in requests-unixsocket shaken, and there are other libraries that do similar things with different protocol names. Switching dependencies would become harder if we exposed `http+unix://` as a supported option.

rockstar:
I tend to agree with you on `os.path.isfile`. I suspect it's use was to fix a possible bug without actually verifying what was actually happening. Please feel free to remove it.

ajkavanagh:
@keesbos What's you status on this? Is it still causing you an issue?  Thanks.

ajkavanagh:
Closing this PR due to no feedback.  Please feel free to re-open it/add an issue if this is still causing you issues.  Thanks.

PR #215, opened 2016-12-22
Handle mode/uid/gid in FilesManager.put()

keesbos:
The FilesManager misses the capability to set file mode/uid/gid

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -79,12 +79,12 @@ class Container(model.Model):
 
         def put(self, filepath, data, mode=None, uid=None, gid=None):
 
-            if isinstance(mode, int):
-                mode = oct(mode)
-            elif not mode.startswith('0'):
-                mode = '0{0}'.format(mode)
             headers = {}
             if mode is not None:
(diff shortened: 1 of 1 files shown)


pylxd/models/container.py
```diff
@@ -77,7 +77,20 @@ def __init__(self, client, container):
             self._client = client
             self._container = container
 
-        def put(self, filepath, data):
+        def put(self, filepath, data, mode=None, uid=None, gid=None):
+
+            if isinstance(mode, int):
```
rockstar: This logic block should fall in under `if mode is not None`.

rockstar:
You've got failing tests, and I think you should add tests for this functionality as well.

ajkavanagh:
@keesbos as @rockstar indicated, this needs tests before it can be merged.  Do you need a hand with that?

ajkavanagh:
I pulled your branch and added tests to it as part of #276 PR.  Thanks for the initial work.  I'm closing this now as I incorporated your commits into PR #276.  Thanks again.

PR #218, opened 2017-02-04
include created_at for snapshot

atrautsch:
It would be nice to have this information.


ajkavanagh:
I'm going to hold off merging this one, as I don't know whether it will break with older versions of LXD.  As soon as I've worked that one out, I'll pull it through.

ajkavanagh:
Okay, sorry for holding so long on this.  It isn't actually backwards compatible, but that's due to a separate bug in pylxd which will get raised shortly.  Thanks very much for the pull request.

PR #220, opened 2017-02-14
fix inhertance of pylxd.models.Container class

om26er:
This code: http://paste.ubuntu.com/23994208/ fails whenever a .sync() is called on a container, the proposed change fixes that.

### DIFF
--- a/pylxd/models/_model.py
+++ b/pylxd/models/_model.py
@@ -152,7 +152,8 @@ class Model(object):
         for key, val in response.json()['metadata'].items():
             if key not in self.__dirty__ or rollback:
                 setattr(self, key, val)
-                self.__dirty__.remove(key)
+                if key in self.__dirty__:
+                    self.__dirty__.remove(key)
         if rollback:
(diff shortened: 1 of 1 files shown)


pylxd/models/_model.py
```diff
@@ -152,7 +152,8 @@ def sync(self, rollback=False):
         for key, val in response.json()['metadata'].items():
             if key not in self.__dirty__ or rollback:
                 setattr(self, key, val)
-                self.__dirty__.remove(key)
+                if key in self.__dirty__:
```
sbaldassin: I think you should move the fix to the if block above. If I got this right, self.__dirty__ will contain all the attributes that need to be sync so if the key is in self.__dirty__ you set the attr and remove the key 
            if key in self.__dirty__ or rollback:
                setattr(self, key, val)
                self.__dirty__.remove(key)

rockstar:
As discussed, we're purposefully not supporting inheritance. I would prefer not to accept this patch at all, as it could hide bugs in the future.

PR #233, opened 2017-06-01
Fix broken CI testing

hsoft:
The whole test suite was broken lately for three reasons:

1. Version in `setp.cfg` was out of sync with git tagging, making `pbr`
   scream.
2. `mock_services` is outdated and unmaintained and doesn't work with
   newer versions of `requests-mock`.
3. The newly added `description` property wasn't properly added in
   mocked LXD.

I've fixed those two problems by updating the `setup.cfg` version,
pinning `requests-mock` to the last version to work and adding a
`description` field to the mocks.

On my machine, all tox tests pass now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
   - test -d .tox/$TOXENV/log && cat .tox/$TOXENV/log/*.log || true
 cache:
   directories:
-    - .tox/$TOXENV
     - $HOME/.cache/pip
 after_success:
   - codecov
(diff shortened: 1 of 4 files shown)


hsoft:
CI checks are failing because in Travis, requests-mock v1.3 was installed. It wasn't supposed to happen because my commits pins it down. I'm now sure why that happened...

hsoft:
Yup, that was travis caches fault. @stgraber do you mind if we remove Travis' tox cache permanently? There's potentially other PRs that are going to get false failures because of that kind of aggressive caching.

stgraber:
@hsoft I don't mind personally. I guess it means slightly longer runs, but it's not like we get a lot of pull request for this project :)

PR #234, opened 2017-06-02
Make models resilient to new attributes in LXD

hsoft:
It's been a few times now (ref #188, #230) where we're in a tricky situation because a new attribute in LXD makes pylxd crash. It's not fun because it makes us *have* to update pylxd and LXD at the same time.

This commit makes models ignore `AttributeError` during `sync()`, which should normally fix this problem once and for all.

Depends on #233

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
   - test -d .tox/$TOXENV/log && cat .tox/$TOXENV/log/*.log || true
 cache:
   directories:
-    - .tox/$TOXENV
     - $HOME/.cache/pip
 after_success:
   - codecov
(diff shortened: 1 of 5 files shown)


hsoft:
@stgraber remember when you said in #188 that it was a bad idea to error on missing attributes? I agree.

hsoft:
Is there anything I can do to help this PR go forward?

robvdl:
Hello @javacruft It seems you do most of the merging on this project.  Would you be so kind to have a look at this commit when you have a spare moment please, it would benefit the lxdock project greatly with doing their next release.

ChrisMacNaughton:
Looks good to me, :+1:

fedordikarev:
For me it looks like this fix is good enough reason to increase version and make new release (2.2.4 or whatever) and we will be happy to have this version available on pypy.python.org

PR #240, opened 2017-07-12
Add support for LXD storage pools

ChrisMacNaughton:
This adds basic support for storage pools as well
as tests to verify their function, taken from the
LXD REST API.

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -45,11 +45,15 @@ class _APINode(object):
             self.session.verify = verify
 
     def __getattr__(self, name):
+        # name here correspoinds to the model name in the LXD API
+        # and, as such, mujsh have underscores replaced with hyphens
         return self.__class__(
             '{}/{}'.format(self._api_endpoint, name.replace('_', '-')),
             cert=self.session.cert, verify=self.session.verify)
 
     def __getitem__(self, item):
+        # item here correspoinds to the model name in the LXD API
+        # and, as such, mujsh have underscores replaced with hyphens
         return self.__class__(
             '{}/{}'.format(self._api_endpoint, item.replace('_', '-')),
             cert=self.session.cert,
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -46,14 +46,14 @@ class _APINode(object):
 
     def __getattr__(self, name):
         # name here correspoinds to the model name in the LXD API
-        # and, as such, mujsh have underscores replaced with hyphens
+        # and, as such, must have underscores replaced with hyphens
         return self.__class__(
             '{}/{}'.format(self._api_endpoint, name.replace('_', '-')),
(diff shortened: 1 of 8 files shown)


pylxd/client.py
```diff
@@ -46,12 +46,12 @@ def __init__(self, api_endpoint, cert=None, verify=True, timeout=None):
...
     def __getattr__(self, name):
         return self.__class__(
-            '{}/{}'.format(self._api_endpoint, name),
+            '{}/{}'.format(self._api_endpoint, name.replace('_', '-')),
             cert=self.session.cert, verify=self.session.verify)
 
     def __getitem__(self, item):
         return self.__class__(
-            '{}/{}'.format(self._api_endpoint, item),
+            '{}/{}'.format(self._api_endpoint, item.replace('_', '-')),
```
ajkavanagh: It's a shame to have to do this, but I can see why it's necessary now.  A comment would be useful for future reference.  (e.g. 'item' == the endpoint model 'e.g. storage-pools, profiles, images')

pylxd/models/storage_pool.py
```diff
@@ -0,0 +1,55 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+from pylxd.models import _model as model
+
+
+class StoragePool(model.Model):
```
ajkavanagh: I think it would be useful to create a class comment to indicate what api endpoint this is expecting to handle, particularly as it is a 'weird' one with a '-' in it.

pylxd/models/storage_pool.py
```diff
@@ -0,0 +1,55 @@
...
+        response = client.api.storage_pools.get()
+
+        storage_pools = []
+        for url in response.json()['metadata']:
+            name = url.split('/')[-1]
+            storage_pools.append(cls(client, name=name))
+        return storage_pools
+
+    @property
+    def api(self):
```
ajkavanagh: Refactoring note for the future.  This method is (nearly) identical in every Model derived class. It's possible other methods here are too.  Not to be changed for the commit though.

pylxd/client.py
```diff
@@ -45,11 +45,15 @@ def __init__(self, api_endpoint, cert=None, verify=True, timeout=None):
             self.session.verify = verify
 
     def __getattr__(self, name):
+        # name here correspoinds to the model name in the LXD API
+        # and, as such, mujsh have underscores replaced with hyphens
```
ajkavanagh: I think this should be 'must' not 'mujsh'?

PR #241, opened 2017-07-20
New: container.execute_interactive()

lingxiaoyang:
Hello there!

This is an experimental implementation of LXD's [interactive `exec`](https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersnameexec). As a contributor of the [lxdock](https://github.com/lxdock/lxdock) project, I would like to add this functionality into pylxd so that we can access a container's shell directly by Python code without having to call `lxc exec` externally. 

The implementation is inspired by [wssh](https://github.com/aluzzardi/wssh/blob/master/wssh/client.py) and you will find similar codes here. I am not good at interactive shell programming but after several trial-and-errors it works! 

To try it out:

````python
import pylxd
c = pylxd.Client()
co = c.containers.get(CONTAINER_NAME)
co.execute_interactive(['/bin/bash'])
````

It should have the exact behavior of `lxc exec CONTAINER_NAME -- /bin/bash`. Resizing the window should send a message to the control websocket and trigger the window resize in the container's shell.

I only tested it on my Ubuntu 16.04, and I don't expect it working on Windows ([refs setup.cfg](https://github.com/lxc/pylxd/blob/master/setup.cfg#L15)).

Regarding interface, another thought could be `container.execute(commands, envs, interactive=True)`. I'll add tests when this is settled.

Your feedback and suggestions are very welcomed!

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -12,12 +12,22 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 import collections
+import fcntl
+import json
+import os
+import platform
+import termios
 import time
+import tty
+import signal
+import struct
+import sys
 
 import six
 from six.moves.urllib import parse
 try:
     from ws4py.client import WebSocketBaseClient
+    from ws4py.client.threadedclient import WebSocketClient
     from ws4py.manager import WebSocketManager
     _ws4py_installed = True
 except ImportError:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


ajkavanagh:
Hi @lingxiaoyang,

Interesting piece of code. Thanks for starting this. I do think that the pylxd library would benefit from an interactive `execute()` type method, but I'm a bit wary of actually exposing/using a terminal *in* the library; that feels like something the application code would.

Certainly #256 **No way of getting command output from container_run_command()** and #244 **No way to pass stdin to Container.execute** both need a similar concept of a STDIN/STDOUT stream to/from the container, although it need not be interactive.  I can also see the interactive need (e.g. supplying 'answers' to scripts in containers, a bit like `expect`).

Perhaps we can build this up, by first doing the STDIN/STDOUT,STDERR mechanism as a sync operation, and then building on that with an ASYNC, threaded, addition? It would then be (relatively) trivial to add on (in a app) the 'tie the STDIN/STDOUT to the actual running terminal' bit.  We could even provide an example of this in an `examples` directory.

ajkavanagh:
Hi @lingxiaoyang, any further thoughts on this?

lingxiaoyang:
Sorry I missed the update of your review @ajkavanagh !

I'll take a look soon. Thank you again for the comments!

lingxiaoyang:
@ajkavanagh  I agree with your opinion. pylxd shouldn't deal with the actual tty.

I reread LXD's API documentation and it's still necessary to separately implement `interactive=False` and `interactive=True` as the numbers of opened websockets are different. For either case, we may experiment SYNC or ASYNC implementation, SYNC more for programmatically respond to stdout/stderr via stdin, ASYNC more for interaction with users. So will there be 4 combinations: interactive+SYNC, interactive+ASYNC, non-interactive+SYNC, non-interactive+ASYNC? (But first, what is the difference between "non-interactive/interactive" and "SYNC/ASYNC"? I'm not very clear of it.)

In the case of SYNC, we may want to add a callback that takes latest stdout+stderr content and gives back the stdin+control content that it wishes to send, most probably on [L399 here in container.py](https://github.com/lxc/pylxd/blob/1df6eff1ea5032353dda8d6a588992783dbcd73c/pylxd/models/container.py#L399). And also, an initial stdin may need to be offered at (I guess, as the API document is not clear) the very moment of establishing the websocket connections.

In the case of ASYNC, we may add a layer of streams (tty and control) as the bus of shared data between pylxd and application code. Can Python io library serve for this purpose (io.IOBase)?

ajkavanagh:
I think that this has now been overtaken by #321.  Could you take a look please; it may be that this PR has extra functionality that would still be useful?

ajkavanagh:
I think we should close this one @lingxiaoyang unless it has extra functionality over #321?

lingxiaoyang:
@ajkavanagh Yes, please go ahead! Thanks for the follow-up.

ajkavanagh:
Thanks very much for doing this; I think it inspired the final solution as well;  please do feel free to contribute again.

ajkavanagh:
Thanks very much for doing this; I think it inspired the final solution as well;  please do feel free to contribute again.

ajkavanagh:
Thanks very much for doing this; I think it inspired the final solution as well;  please do feel free to contribute again.

PR #245, opened 2017-09-21
Fix memory hog during image import

shurkaxaa:
Existing version load image into memory when non-unified image used
(with separate tarball with manifest).
Added ability for stream based image upload without loading them to memory.
requests-toolbelt external package required for streaming multipart HTTP
image upload.
Additional parameter added to identify if streams are/can be passed - this will allow
pylxd client to identify via introspection if pylxd library supports stream based
upload and fallback to old non-stream behavior if old version of pylxd is installed.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -4,32 +4,33 @@ Contributors to pylxd
 
 These are the contributors to pylxd according to the Github repository.
 
- =============== ==================================
- GH username      Name
- =============== ==================================
- rockstar        Paul Hummer
- zulcss          Chuck Short
- saviq           Michał Sawicz
- javacruft       James Page (Canonical)
- pcdummy         Rene Jochum
- jpic            ???
- hsoft           Virgil Dupras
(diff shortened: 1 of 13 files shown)


ajkavanagh:
I think the tests are failing because your branch is behind master.  Please could you rebase your branch against pylxd/master and we can see if the tests will pass.  Thanks.

shurkaxaa:
Rebased as requested. Now CI tests passed.

ajkavanagh:
Hi.  It looks like something is wrong with the PR from a rebasing perspective.  We're going to need to fix this PR before it can merge because it's showing commits that were made on master in your PR.  Did you do a merge rather than a rebase?  What we would like in the PR is just the changes that the PR wants to make to the master branch, which I think are quite small.  It might be easiest to fix this, but closing the PR (but make sure you don't lose your work) and, re-branching from master, and re-applying your changes and doing a new PR?  Thanks.

shurkaxaa:
Ok, new pull request created #252. This PR should be closed.

PR #247, opened 2017-09-26
Adjust in the function snapshot.all()

shazaum:
Use clearer function:

snap =  client.snapshot.all()
for snaps in snap:
   print snaps.name

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -237,6 +237,7 @@ class Client(object):
         self.networks = managers.NetworkManager(self)
         self.operations = managers.OperationManager(self)
         self.profiles = managers.ProfileManager(self)
+        self.snapshot = managers.SnapshotManager(self)
 
     @property
     def trusted(self):
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -384,13 +384,18 @@ class Snapshot(model.Model):
         return snapshot
 
     @classmethod
-    def all(cls, client, container):
-        response = client.api.containers[container.name].snapshots.get()
 
-        return [cls(
-                client, name=snapshot.split('/')[-1],
-                container=container)
-                for snapshot in response.json()['metadata']]
+    def all(cls, client):
+        response = client.api.containers.get()
+        containers = []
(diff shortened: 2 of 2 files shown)


ajkavanagh:
@shazaum thanks for the PR.  Unfortunately, I don't understand what this is a) fixing or b) what the new feature is?

Please could you indicate if this is a fix to a bug report (and quote the bug report) or describe what the feature is more clearly, and why the existing implementation is inadequate.  Thanks.

shazaum:
@ajkavanagh Ops...
What I did was list the snapshots of all the containers directly.
It is not to replace the current function.
Just one way to list all snapshots directly.

shazaum:
Just a new feature.

ajkavanagh:
@shazaum thanks for the clarification.  In this situation, unfortunately, this patch doesn't fit with the design of the library's API, and also breaks the existing API call.  So it can't currently be accepted.

The issues are:

1. It breaks the existing implementation of the `...snapshot.all()`.
2. It doesn't really map to a single API call.
3. It's fairly easy to implement as a function that calls the library.

One way to do this using the API is:

```python
client = pylxd.Client()
snapshots = []
for container in client.containers.all():
    snapshots.extend(container.snapshots.all())
print(", ".join(s.name for s in snapshots))
```

This will print the names of the snapshots if you snapshot them with a name that is useful:

```bash
lxc snapshot <container-name> <snapshot-name>
```

shazaum:
I think the lib can deliver a function to show all the snapshots...
because you first have to say, I want all the containers, then ask for the snapshots.

ajkavanagh:
I'm sorry but this isn't going to be added as the LXD RESTful API doesn't provide the an API to do this directly.  Exactly the same issue of 1. getting all the containers, 2. iterating through the containers to get the snapshots, would have to be implemented in the library.

i.e. there is no `client.snapshots.all()` equivalent API call, and so the library would have to manufacture it using the code noted above.  This isn't in the spirit/design of the library which is to map as closely as possible to the RESTful API at: https://github.com/lxc/lxd/blob/master/doc/rest-api.md

PR #252, opened 2017-09-27
Image import - memory hog fix

shurkaxaa:
Existing version load image into memory when non-unified image used
(with separate tarball with manifest).
Added ability for stream based image upload without loading them to memory.
requests-toolbelt external package required for streaming multipart HTTP
image upload.

### DIFF
--- a/pylxd/models/image.py
+++ b/pylxd/models/image.py
@@ -122,7 +122,7 @@ class Image(model.Model):
         if public:
             headers['X-LXD-Public'] = '1'
 
-        if from_streams is not None:
+        if from_streams:
             # Image uploaded as chunked/stream (metadata, rootfs)
             # multipart message.
             # Order of parts is important metadata should be passed first
--- a/pylxd/models/image.py
+++ b/pylxd/models/image.py
@@ -11,10 +11,12 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+import collections
 import contextlib
 import tempfile
 import uuid
 import warnings
+from requests_toolbelt import MultipartEncoder
 
 import six
 
@@ -99,7 +101,8 @@ class Image(model.Model):
 
     @classmethod
     def create(
-            cls, client, image_data, metadata=None, public=False, wait=True):
+            cls, client, image_data,
+            metadata=None, public=False, wait=True, from_streams=False):
         """Create an image.
 
         If metadata is provided, a multipart form data request is formed to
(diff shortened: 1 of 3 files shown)


pylxd/models/image.py
```diff
@@ -119,7 +122,17 @@ def create(
         if public:
             headers['X-LXD-Public'] = '1'
 
-        if metadata is not None:
+        if from_streams:
```
ChrisMacNaughton: Is there a compelling reason that we'd ever want to _not_ stream the image? If not, I'd prefer that we not add a new option and just chunk all image uploads

shurkaxaa: I wrote it in comment - the only reason if we want to have pylxd clients to work with old version too.
For example nova-lxd - if it will be switched to stream based image upload it will require NEW version of pyxld i.e it will not be able to wold with old version of pylxd. I do not see any versioning in pylxd code so client will not be able which version of pyxld is installed.
As I was not sure that pyxld and nova-lxd will be updated synchronously by packagers I assume that clients can fallback to non-stream behavior via introspection of create method.
In short it will allow client of pyxld library to work with both version of pyxld - fixed one and old one. If we are sure that pyxld will be ALWAYS updated as it's client we can remove this for sure and assume that streams always passed to create operation.

shurkaxaa: May be it make sense at least to dynamically identify if streams are passed or content as before and act accordingly?
It will provide backward compatibility for 'old' clients which will pass content instead of streams. What do you think?

pylxd/tests/models/test_image.py
```diff
@@ -110,7 +111,8 @@ def test_create_with_metadata(self):
         """An image with metadata is created."""
         fingerprint = hashlib.sha256(b'').hexdigest()
         a_image = models.Image.create(
-            self.client, b'', metadata=b'', public=True, wait=True)
+            self.client, StringIO(u''), metadata=StringIO(u''),
```
ChrisMacNaughton: Instead of editing this test, can you please add another test that verifies the new functionality while allowing us to maintain a test to ensure no future regressions in the API handling?

shurkaxaa:
Old functionality no longer available if I will keep old test it will fail. You ask me before to always assume that stream based upload used, so modified test no longer relevant as it does not use streams. If I will keep it as is it will fail.

shurkaxaa:
'Old' test returned back as requested.
I've checked that requests_toolbelt multipart encoder
accept both types on input - binary string and file like object.
It means that backward compatibility with old clients guaranteed.
So old clients will still upload images as binary strings and new one can use streams.

ChrisMacNaughton:
@ajkavanagh Thoughts?

PR #253, opened 2017-09-27
Update cert check to look in snap path as well

### LINKED ISSUE #251: snap installed LXD puts client certificates in a new location
The new path for the crt and key files is in `~/snap/lxd/current/.config/lxc/` rather than the previous default of `~/.config/lxc` so pylxd should be able to detect the correct one and use those

ChrisMacNaughton:
When using a snap installed LXD, we need to look in the $HOME/snap
directory rather than looking in ~ for the certificates

Fixes #251

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -33,15 +33,16 @@ requests_unixsocket.monkeypatch()
 LXD_PATH = '.config/lxc/'
 SNAP_ROOT = '~/snap/lxd/current/'
 APT_ROOT = '~/'
-if os.path.exists(os.path.expanduser(SNAP_ROOT)):
-    CERTS_PATH = os.path.join(SNAP_ROOT, LXD_PATH)
-else:
-    CERTS_PATH = os.path.join(APT_ROOT, LXD_PATH)
+if os.path.exists(os.path.expanduser(SNAP_ROOT)):  # pragma: no cover
+    CERTS_PATH = os.path.join(SNAP_ROOT, LXD_PATH)  # pragma: no cover
+else:  # pragma: no cover
+    CERTS_PATH = os.path.join(APT_ROOT, LXD_PATH)  # pragma: no cover
 
-Cert = namedtuple('Cert', ['cert', 'key'])
+Cert = namedtuple('Cert', ['cert', 'key'])  # pragma: no cover
 DEFAULT_CERTS = Cert(
     cert=os.path.expanduser(os.path.join(CERTS_PATH, 'client.crt')),
-    key=os.path.expanduser(os.path.join(CERTS_PATH, 'client.key')))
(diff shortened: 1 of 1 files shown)


pylxd/client.py
```diff
@@ -29,6 +29,17 @@
 
 requests_unixsocket.monkeypatch()
 
+LXD_PATH = '.config/lxc/'
+SNAP_ROOT = '~/snap/lxd/current/'
+APT_ROOT = '~/'
+if os.path.exists(os.path.expanduser(SNAP_ROOT)):
+    CERTS_PATH = "{}{}".format(SNAP_ROOT, LXD_PATH)
```
ajkavanagh: I'd like to see this use the `os.path.join()`: e.g.:

```python
CERT_PATH=os.path.join(SNAP_ROOT, LXD_PATH)
```

pylxd/client.py
```diff
@@ -29,6 +29,17 @@
...
+else:
+    CERTS_PATH = "{}{}".format(APT_ROOT, LXD_PATH)
+DEFAULT_CERTS = (
+    os.path.expanduser('{}client.crt'.format(CERTS_PATH)),
+    os.path.expanduser('{}client.key'.format(CERTS_PATH)))
```
ajkavanagh: And here too, please.  It will stop bizarre errors in the future when someone edits the path and leaves a separator off.

pylxd/client.py
```diff
@@ -214,9 +216,9 @@ def __init__(
                 # Extra trailing slashes cause LXD to 301
                 endpoint = endpoint.rstrip('/')
                 if cert is None and (
-                        os.path.exists(self.DEFAULT_CERTS[0]) and
-                        os.path.exists(self.DEFAULT_CERTS[1])):
-                    cert = self.DEFAULT_CERTS
+                        os.path.exists(DEFAULT_CERTS[0]) and
+                        os.path.exists(DEFAULT_CERTS[1])):
```
ajkavanagh: I think we have an opportunity to improve the readability of the code by using a `namedtuple()`.  See [collections.namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) for details.  Basically, instead of `DEFAULT_CERTS[1]` it would be possible to write `DEFAULT_CERTS.cert` and `DEFAULT_CERTS.key`.  Ask me if you need further details.

pylxd/client.py
```diff
@@ -210,6 +210,23 @@ class Client(object):
     def __init__(
             self, endpoint=None, version='1.0', cert=None, verify=True,
             timeout=None):
+        """Constructs a LXD client
+
+        :param endpoint: (optional): adnpoint can be an http endpoint or
```
ajkavanagh: Typo here?  "adnpoint@

PR #262, opened 2018-01-03
Add FilesManager.delete and corresponding test

### LINKED ISSUE #261: FilesManager.delete is missing
This is the code that handles the container `/files` endpoint:

https://github.com/lxc/pylxd/blob/master/pylxd/models/container.py#L74-L91

It is missing `delete`. This should be very simple to implement since all that needs to be done is to call into the api.

I will create a PR for this soon - please let me know if there's anything I need to consider.

duk3luk3:
Fixes #261


pylxd/models/container.py
```diff
@@ -84,6 +84,12 @@ def put(self, filepath, data):
                 params={'path': filepath}, data=data)
             return response.status_code == 200
 
+        def delete(self, filepath):
+            response = self._client.api.containers[
```
ajkavanagh: We probably want to put a test in as to whether the feature is supported.

e.g.

```python
if u'file_delete' not in self._client.host_info['api_extensions']:
    raise some error that makes sense here.
```

On one of my test boxes (16.04 LTS for example), my lxd is 2.0.8 which is missing that feature.  This would save a round-trip to the server when it's not going to work.

See  https://github.com/lxc/lxd/blob/master/doc/api-extensions.md and https://github.com/lxc/lxd/blob/master/doc/rest-api.md#delete-pathpathinsidethecontainer for more details.

pylxd/tests/models/test_container.py
```diff
@@ -431,13 +431,15 @@ def setUp(self):
         super(TestFiles, self).setUp()
         self.container = models.Container.get(self.client, 'an-container')
 
-    def test_put(self):
-        """A file is put on the container."""
+    def test_put_delete(self):
```
ajkavanagh: Would also need to add a test for whether the api_extension is supported and handle the negative/raise, if it's not.

pylxd/models/container.py
duk3luk3: Aha, that explains why it was missing.

pylxd/tests/models/test_container.py
duk3luk3: The tests are currently running against an entirely mocked LXD API as far as I can tell. I'll have to figure out how to flip the extension support.

duk3luk3:
This commit now introduces some pretty janky things to the test setup, including changing the server definition endpoint response mock. It's hard to feel bad about that though because the completely mocked server doesn't seem like a very natural or faithful test target to me.

The code I've come up with might not be the right way to do this, but I didn't realize that once I started and then decided to see it through to the end to see where it would take me.

This now makes the tests a good bit more fragile that might break due to unrelated changes or, worse, might mask any changes made to the server definition endpoint, potentially creating frustration for anyone trying to change the definition later.

I see two ways to fix that:
* Create another mocked server "instance" that has the api_extension defined
* Change `PyLXDTestCase.add_rule` class so that rule changes can be reverted

I personally would prefer the second choice. Adding another mocked server instance feels like creating more complexity than can be easily handled by the current test setup without refactoring the mock setup to create clearly delineated "server test instances" with different behaviour. Creating a way to push and pop rule updates allows to keep one static set of basic rules for the server while keeping any modifications for specific test cases inside those test cases, making the complexity manageable because changes are close to what they are needed for.

pylxd/tests/testing.py
```diff
@@ -35,3 +35,6 @@ def tearDown(self):
     def add_rule(self, rule):
         """Add a rule to the mock LXD service."""
         mock_services.update_http_rules([rule])
+
```
ajkavanagh: This `is_mocked()` isn't required as the `def setUp()` above calls `mock_services.start_http_mock()` which means that when the test runs and calls `is_mocked()` it would always be true.  No other test seems to need it?

pylxd/tests/models/test_container.py
```diff
@@ -431,13 +431,39 @@ def setUp(self):
...
         data = 'The quick brown fox'
 
-        self.container.files.put('/tmp/putted', data)
-
-        # TODO: Add an assertion here
+        res = self.container.files.put('/tmp/putted', data)
+        self.assertEqual(True, res, msg='Failed to put file, result: {}'.format(res)) # NOQA
+
+        # if we are mocked, delete should initially not be available
+        if self.is_mocked():
```
ajkavanagh: Shouldn't need this `if self.is_mocked()` test; the service will be mocked as the test case is derived from `...testing.PyLXDTestCase`.

ajkavanagh:
What you've done seems reasonable.  Mocking out the service comes with the problem of keeping the mocks up to date with the actual service.  It would be cool if we could automagically parse the service definition from LXD, but that's probably a job for another day.  However, this part of the API should be pretty stable (as should the whole API) with only stuff being added rather than fundamentally changed.  So it should be on reasonable ground.

Thanks again for working on this.

pylxd/tests/testing.py
duk3luk3: I'm adding `is_mocked` so that the test suite could also be run against a real LXD. Is that not a goal?

ajkavanagh: Nope, there's no need to run the tests against a real LXD.  These are unit tests as opposed to integration tests.  There are some integration tests, but we probably need a better story on it.  That's going to be one of the goals going forward; it is used (partially) in nova-lxd (part of OpenStack) and gets some integration tests there, but doesn't have anything like a full integration test.

Equally, it is just converting 'python calls' into REST-like API calls; and these are checked in the mocks.  So that's not too bad.

In summary, no the `is_mocked()` isn't needed.

duk3luk3: OK, but you don't see a problem with me mangling the mock to change the api_extensions as I am doing?

duk3luk3:
I removed the mocking checks as requested.

PR #265, opened 2018-01-10
Implemented recursive file transfer

mrtc0:
https://github.com/lxc/pylxd/issues/246

`Container.files.recursive_put(src, dst)`, can recursive transfer of files.
This means that folder transfer is possible.

```bash
$ lxc exec c1 -- ls -al test
total 8
drwxr-xr-x    2 root     root          4096 Jan 10 14:58 .
drwx------    3 root     root          4096 Jan 10 14:58 ..
$ ls doc/source
api.rst            authentication.rst certificates.rst   conf.py            containers.rst     contributing.rst   events.rst         images.rst         index.rst          installation.rst   networks.rst       profiles.rst       usage.rst
```

```python
...
In [7]: c1.files.recursive_put('./doc', '/root/test')
Out[7]: True
```

```bash
$ lxc exec c1 -- ls test/doc
source
$ lxc exec c1 -- ls test/doc/source
api.rst             authentication.rst  certificates.rst    conf.py             containers.rst      contributing.rst    events.rst          images.rst          index.rst           installation.rst    networks.rst        profiles.rst        usage.rst
```


pylxd/models/container.py
```diff
@@ -90,6 +91,27 @@ def get(self, filepath):
...
 
+        def recursive_put(self, src, dst):
+            if os.path.isdir(src):
+                idx = len(os.path.dirname(src))
+                dst_dirs = []
+                for path, dirname, files in os.walk(src):
+                    dst_dir = os.path.join(dst, path[idx:].lstrip(os.path.sep))
+                    dst_dirs.append(dst_dir)
+
+                self._container.execute(['mkdir', '-p'] + dst_dirs)
```
ajkavanagh: `dst_dirs` might end up being very long if there are many directories and also lots of duplicate parents if it is deeply nested.  What's really needed here is just to know the "tips" that need to be created, rather than all of the intermediates.  Using a `collections.defaultdict(dict)` and then just using values that are not dicts would cut down on the line length.

Alternatively, make the directories as you go by just traversing once, but using the API to push a directory rather than a file.  (mod to file put needed)

ajkavanagh:
Thanks for starting this piece of work.  I've done a quick review, but I also wanted to offer a few suggestions for improving it to get it to a mergeable state:

1. Can we eliminate one of the `os.walk(...)` loops? e.g. the code code do the `mkdir -p` on each path from the outer `for` loop on os.walk(...).
2. Speaking of `mkdir -p`, the LXD API for [POST /1.0/containers/<name>/files?path=/path/inside/the/container](https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-pathpathinsidethecontainer) has a HEADER option of `X-LXD-type` which can specify a directory.  So we could eliminate the `mkdir` exec call and use the API directly to create the path; obviously it *might* fail, so we would effectively be ignoring the return value.  If LXD introduce a 'stat' like API call, we could stat it first to make sure that it was a file or directory to match what we're trying to do.
3. The put API has options for the `uid` and `gid`; it might be nice to add that to the pylxd function as defaults (=0) and set them on the headers for the post.
4. The code won't merge without unit tests.

Let me know if this is something that you will have the time to do; if not, we'll find someway of getting it done!.  Thanks again.

ajkavanagh:
@mrtc0 hi -- are you planning to do any more on this soon, please?  I'd like to plan it for 2.2.5 (next couple of weeks) or 2.2.6 within the next 3 months if possible.  Thanks.

mrtc0:
@ajkavanagh I’m sorry for the delay. I'll work on it soon.

ajkavanagh:
Hi @mrtc0 sorry, didn't mean to increase urgency on you; just wanted to be able to plan it in - no pressure to fix it now, unless you'd personally like it to go into 2.2.5 (soon) or 2.2.6 by April, or later.  Thanks again.

ajkavanagh:
@mrtc0 just to let you know that we've merged #276 which adds `uid`, `gid` and `mode` to the single file `put` command.  It may be relevant to the work here.

pylxd/models/container.py
```diff
@@ -143,25 +143,52 @@ def get(self, filepath):
...
+
+            :param src: The source path of directory to copy.
+            :type filepath: str
+            :param dst: The destination path in the container
+                    of directory to copy
+            :type filepath: str
+            """
+            norm_src = os.path.normpath(src)
+            if not os.path.isdir(norm_src):
+                raise NotADirectoryError(
```
ajkavanagh: This will fail on py2.7 -- `NotADirectoryError` only came into being with Py3.4 (IIRC!).

We either need to drop it to a `ValueError`or create a `NotADirectoryError` Exception class here, by subclassing `Exception` if we're in Python2.7.  Thoughts?

pylxd/tests/models/test_container.py
```diff
@@ -431,6 +433,16 @@ def setUp(self):
         super(TestFiles, self).setUp()
         self.container = models.Container.get(self.client, 'an-container')
 
+    @classmethod
+    def setUpClass(TestFiles):
```
ajkavanagh: Putting the temp files in the `setUpClass()` method is probably making the test class do "too much work" as every test in the class will create and tear down the files regardless of whether they are needed.  I suggest that they are moved to the test method, please.

pylxd/tests/models/test_container.py
```diff
@@ -514,6 +526,35 @@ def capture(request, context):
         with self.assertRaises(ValueError):
             res = self.container.files.put('/tmp/putted', data, mode=object)
 
+    def test_recursive_put(self):
+        """ Recursive put directory to the container """
```
ajkavanagh: This test doesn't check that the header is actually set for directory when creating a directory.  Something also the lines of https://github.com/lxc/pylxd/blob/master/pylxd/tests/models/test_container.py#L471 where the headers are 'captured' during the test and can be checked would be very useful.

ajkavanagh:
I've done some work with this and I've done a PR to @mrtc0 for further conversation ... see https://github.com/mrtc0/pylxd/pull/1

ajkavanagh:
Hi @mrtc0 thanks for the updates;  just one tiny failure and we'd be good to go on pep8, please:

```
pep8 runtests: commands[0] | flake8 --ignore=E123,E125
./pylxd/exceptions.py:42:69: W292 no newline at end of file
        """ An exception raised when not a directory for python2 """                                                                    ^
ERROR: InvocationError for command '/home/travis/build/lxc/pylxd/.tox/pep8/bin/flake8 --ignore=E123,E125' (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   pep8: commands failed
```

ajkavanagh:
Thanks very much @mrtc0 for working through the changes; this is now great to go.

ajkavanagh:
@mrtc0 If you'd like to, you can add yourself to the `CONTRIBUTORS.rst` via another pull request.  Thanks again for implementing the feature.

PR #267, opened 2018-01-23
Add used_by to profile attributes

e1ee1e11:
Hi, 
I fix typos in profiles.rst when reading the documentation.
After that, I add "used_by" to profile attributes and arrange attributes in alphabetical order .

### DIFF
--- a/pylxd/models/profile.py
+++ b/pylxd/models/profile.py
@@ -17,10 +17,11 @@ from pylxd.models import _model as model
 class Profile(model.Model):
     """A LXD profile."""
 
-    name = model.Attribute(readonly=True)
-    description = model.Attribute()
     config = model.Attribute()
+    description = model.Attribute()
     devices = model.Attribute()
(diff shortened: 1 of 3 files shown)


pylxd/models/profile.py
```diff
@@ -17,10 +17,11 @@
 class Profile(model.Model):
     """A LXD profile."""
 
-    name = model.Attribute(readonly=True)
-    description = model.Attribute()
     config = model.Attribute()
+    description = model.Attribute()
     devices = model.Attribute()
+    name = model.Attribute(readonly=True)
+    used_by = model.Attribute(readonly=True)
```
ajkavanagh: It would be a good idea to add `'used_by': [],` to `def profile_GET(...)` in the `pylxd/tests/mock_lxd.py` file to ensure the test cases match what's coming back.

ajkavanagh:
@e1ee1e11 Thanks very much for the PR.  Now merged.

PR #269, opened 2018-01-24
Update the containers documentation for 2.2.5

### LINKED ISSUE #260: Documentation is outdated
Happy new year! 🎉 

The documentation is in pretty bad shape. Example:

https://pylxd.readthedocs.io/en/latest/containers.html / https://github.com/lxc/pylxd/blob/master/doc/source/containers.rst:

> execute - Execute a command on the container. The first argument is a list, in the form of subprocess.Popen with each item of the command as a separate item in the list. Returns a two part tuple of (stdout, stderr). This method will block while the command is executed.

`container.py`:

https://github.com/lxc/pylxd/blob/8588c3afb316299f204ae502b1c35dd7d7e3420e/pylxd/models/container.py#L39-L41

https://github.com/lxc/pylxd/blob/8588c3afb316299f204ae502b1c35dd7d7e3420e/pylxd/models/container.py#L250-L251

Also:

https://github.com/lxc/pylxd/blob/8588c3afb316299f204ae502b1c35dd7d7e3420e/pylxd/models/container.py#L207-L212

Why is the documentation in the `doc` folder instead of inline? Why is even the inline documentation becoming outdated?

LXC is an amazing project (really!) and it is a shame that the documentation is not excellent.

Python documentation is not something I've done a lot of work with, but if anyone commits to making the technical process to generate docs and push them to readthedocs work, I would be willing to spend effort on moving the documentation from `doc/` to docstrings in the source and cleaning it up in the process.

### LINKED ISSUE #263: Image creation from container is undocumented
Is it correct that there is no implementation for creating an image from a container, as in the "source container" case described here: https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-7 ?

ajkavanagh:
Update the documentation to match the current functionality offered by the library.
Fixes to docs to remove warnings/errors from sphinx autodoc class documentation generation.
Additions of several pages to cover new functionality.

Closes #260 
Closes #263

### DIFF
--- a/doc/source/storage-pools.rst
+++ b/doc/source/storage-pools.rst
@@ -1,0 +1,44 @@
+Storage Pools
+=============
+
+LXD supports creating and managing storage pools and storage volumes. General
+keys are top-level. Driver specific keys are namespaced by driver name. Volume
+keys apply to any volume created in the pool unless the value is overridden on
+a per-volume basis.
+
+`Storage Pool` objects are represent the dictionary and associated methods that
+are returned from the API endpoint `GET /1.0/storage-pools/<name>`.
+
+Manager methods
+---------------
+
+Storage-pools can be queried through the following client manager methods:
+
+  - `all()` - Return a list of storage pools.
+  - `get()` - Get a specific storage-pool, by its name.
+  - `exists()` - Return a boolean for whether a storage-pool exists by name.
+  - `create()` - Create a storage-pool.  **Note the config in the create class
+    method is the WHOLE json object described as `input` in the API docs.**
+    e.g. the 'config' key in the API docs would actually be `config.config` as
+    passed to this method.
+
+
+Storage-pool attributes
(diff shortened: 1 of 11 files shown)


doc/source/storage-pools.rst
```diff
@@ -0,0 +1,44 @@
+Storage Pools
+=============
+
+LXD supports creating and managing storage pools and storage volumes. General
+keys are top-level. Driver specific keys are namespaced by driver name. Volume
+keys apply to any volume created in the pool unless the value is overridden on
+a per-volume basis.
+
+`Storage Pool` objects are represent the dictionary and associated methods that
```
ChrisMacNaughton: `objects are represent the dictionary` - probably want to edit to `objects are represented by the dictionary`

ajkavanagh: Good catch :)

PR #283, opened 2018-03-07
Add proposal for version 3.0 of the pylxd API

ajkavanagh:
This is a proposal/spec to explain what a v3 of the pylxd API would be
like and why it is needed.  This would be a breaking change for clients
of pylxd.

It's opened for discussion here; thoughts both for and against are welcomed.

### DIFF
--- a/notes/version-3-proposal.rst
+++ b/notes/version-3-proposal.rst
@@ -36,8 +36,8 @@ purpose of the library.
 
 2. The attribute system only tries to do a single level of JSON loading, and
    doesn't load the whole tree of metadata returned in an API call.  Thus you
-   get the (slightly) perverse situation of some attributes container JSON
-   encoded strings, rather than a dictionary/list object.  See `#214
+   get the (slightly) perverse situation of some attributes container contain
+   JSON encoded strings, rather than a dictionary/list object.  See `#214
    <https://github.com/lxc/pylxd/issues/214>`_ for example.
 
 3. Each ``object`` returned has hard-coded methods for ``.api``, ``.client``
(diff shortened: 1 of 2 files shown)


notes/version-3-proposal.rst
```diff
@@ -0,0 +1,326 @@
...
+   c = pylxd.Client()
+   profile = c.containers['test'].profiles[0]
+
+and the ``profile`` object would be a pylxd ``Model()`` object, where the
+``used_by`` attribute would be a dictionary like object which had 'test' as a
+key, and the test container as a value. e.g.
+
+.. code-block:: python
+
+   container = c.profiles['default'].used_by
```
ChrisMacNaughton: Should this be `containers`?

notes/version-3-proposal.rst
```diff
@@ -0,0 +1,326 @@
...
+   <... Manager for containers ...>
+
+i.e. the ``used_by`` property would return a manager to access profiles.
```
ChrisMacNaughton: I believe that this manager would access containers?

notes/version-3-proposal.rst
```diff
@@ -0,0 +1,326 @@
...
+addressed to give a more natural Pythonic interface to lxd, which is the
+purpose of the library.
+
+1. The attribute/manager/__slots__ system, whilst very clever, is inflexible
+   and simply discards new attributes that are added to (point) releases of the
+   LXD package.
+
+2. The attribute system only tries to do a single level of JSON loading, and
+   doesn't load the whole tree of metadata returned in an API call.  Thus you
+   get the (slightly) perverse situation of some attributes container JSON
```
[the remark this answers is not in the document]
ajkavanagh: Yes, it should.  Thanks for the catch.

notes/version-3-proposal.rst
ajkavanagh: IIRC,  `c` at this point is a container object, and thus this would be correct.  I'll check though.

notes/version-3-proposal.rst
ajkavanagh: And, yes, this is true too.

notes/version-3-proposal.rst
ChrisMacNaughton: I understand, and concur, that `c` is a container; I'm concerned that, on asking for the profile, that we would not then have the list of containers using that profile (ex: default has many containers)

ajkavanagh: I think what I may not have explained well is that `used_by` would be a `ContainerManager` instance, meaning that it represents 0..n containers.  I'll change the initial variable assignment to `container_manager` and add some words.

PR #289, opened 2018-03-19
Bug/280/long running container

ajkavanagh:
The bug presented as the operation not still existing at the lxd
daemon when the end of the command was reached.  After much testing,
it seemed that the pylxd client was closing the websockets when the
empty message was received on them, which turned out to be too late
for some kinds of output.  Instead, tracking when the operation
completed, and then waiting for the sockets to close meant that the
return code of the command could be obtained, and the wait for the
sockets to be closed by the lxd daemon ensured completion.  The code
also waits for the ws4py manager to clean up and waits on the threads to
end.  This helps with multiple commands.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,5 +2,6 @@ include AUTHORS
 include ChangeLog
 exclude .gitignore
 exclude .gitreview
+exclude contrib_testing
 
 global-exclude *.pyc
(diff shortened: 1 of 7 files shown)


ajkavanagh:
The integration tests in contrib demonsrate the fix.  It's hard to do it programmatically because it needs network type traffic to expose the race.

PR #296, opened 2018-04-12
List operations and cancel operations added features

lasizoillo:
Signed-off-by: javier.lasheras <javier.lasheras@avature.net>

Related to 271 issue. Websocket functionality need more work.

### DIFF
--- a/pylxd/models/operation.py
+++ b/pylxd/models/operation.py
@@ -40,8 +40,28 @@ class Operation(object):
         response = client.api.operations[operation_id].get()
         return cls(_client=client, **response.json()['metadata'])
 
-    def __init__(self, **kwargs):
+    @classmethod
+    def all(cls, client):
+        """List of operations"""
+        response = client.api.operations.get()
+
+        operations = []
+        for op_id in response.json()['metadata']['running']:
+            operations.append(cls.get(client, op_id))
+        return operations
+
+    @property
+    def api(self):
+        return self._client.api.operations[self.id]
+
+    @property
+    def class_(self):
+        """ Avoid class as reserved keyworkds """
+        return getattr(self, "class")
+
+    def __init__(self, client=None, **kwargs):
         super(Operation, self).__init__()
+        self._client = client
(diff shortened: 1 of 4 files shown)


pylxd/models/operation.py
```diff
@@ -40,8 +40,28 @@ def get(cls, client, operation_id):
...
+            operations.append(cls.get(client, op_id))
+        return operations
+
+    @property
+    def api(self):
+        return self._client.api.operations[self.id]
+
+    @property
+    def class_(self):
+        """ Avoid class as reserved keyworkds """
```
ajkavanagh: There's a typo in the docstring: "keyworkds" -> "keywords"

Also, the code coverage is stying that the `return ...` is never tested and thus throwing it out.  Please could you adjust the tests just to exercise this (admittedly tiny) bit of coverage?

pylxd/models/operation.py
```diff
@@ -40,8 +40,28 @@ def get(cls, client, operation_id):
...
+        return getattr(self, "class")
+
+    def __init__(self, client=None, **kwargs):
```
ajkavanagh: You'll need to re-base your PR, please, as there was a bug fixed in the `operation.py` file; please see the master branch.

pylxd/models/operation.py
```diff
@@ -64,3 +84,15 @@ def wait(self):
...
+        """
+        Cancel an operation. If may_cancel is False do nothing.
+
+        Calling this will change the state to "cancelling" rather than
+        actually removing the entry.
+        """
+        if not self.may_cancel:
+            return
+        self.api.delete()
+        self.status = "cancelling"
```
ajkavanagh: We shouldn't set `self.status = "cancelling"` as it might not reflect the corresponding state in the object in lxd.
What we should do here is fetch the operation again, and copy the `status`, `status_code`, `may_cancel` and `err` attributes back onto the class.

Alternatively, we should promote it to a `_Model` and just call `sync()` which might be simpler, overall.

Issue #302, opened 2018-04-24
Problem with integration testing


ppkt:
Hi,
I started implementing API for Network - I created few unit tests and everything was working quite good but when I started working on integration tests I'm getting 'not implemented' exception when I'm trying to create new network:
```
(integration) root@lxd-test:/opt/pylxd# python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pylxd import Client
>>> c = Client()
>>> c.networks.all()
[<pylxd.models.network.Network object at 0x7f49fa1e8840>, <pylxd.models.network.Network object at 0x7f49fa1e88b8>, <pylxd.models.network.Network object at 0x7f49fa1e8930>]
>>> c.networks.create(name='test')
<pylxd.client._APINode object at 0x7f49fa5454a8>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/pylxd/pylxd/models/network.py", line 67, in create
    client.api.networks.post(json=network)
  File "/opt/pylxd/pylxd/client.py", line 157, in post
    self._assert_response(response, allowed_status_codes=(200, 201, 202))
  File "/opt/pylxd/pylxd/client.py", line 104, in _assert_response
    raise exceptions.LXDAPIException(response)
pylxd.exceptions.LXDAPIException: not implemented
```

But when I run the same code on my local machine (not in container) everything is working correctly:

```
$ python3
Python 3.6.5rc1 (default, Mar 14 2018, 06:54:23) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pylxd import Client
>>> c = Client()
>>> net = c.networks.create(name='test')
<pylxd.client._APINode object at 0x7fd87d76ee48>
/home/ppkt/workspace/pylxd/pylxd/models/_model.py:116: UserWarning: Attempted to set unknown attribute "status" on instance of "Network"
  key, self.__class__.__name__
/home/ppkt/workspace/pylxd/pylxd/models/_model.py:116: UserWarning: Attempted to set unknown attribute "locations" on instance of "Network"
  key, self.__class__.__name__
>>> net.marshall()
[...]

PR #303, opened 2018-04-25
Add network operations

ppkt:
I added support for operations on networks. I based my code on Profile model because it was quite similar. However, due to issue #302 it's not possible to test my changes using 'run_integration_tests' script on Ubuntu 16.04 and after switching to 17.10 my tests are passing but few other are failing (due to missing storage pool).

### DIFF
--- a/doc/source/networks.rst
+++ b/doc/source/networks.rst
@@ -1,7 +1,11 @@
+.. py:currentmodule:: pylxd.models
+
 Networks
 ========
 
-`Network` objects show the current networks available to lxd.
+:class:`Network` objects show the current networks available to LXD. Creation
+and / or modification of networks is possible only if 'network' LXD API
+extension is present (see :func:`~Network.network_extension_available`)
 
 
 Manager methods
@@ -10,11 +14,13 @@ Manager methods
 Networks can be queried through the following client manager
 methods:
 
-  - `all()` - Retrieve all networks.
-  - `exists()` - See if a profile with a name exists.  Returns `boolean`.
-  - `get()` - Get a specific network, by its name.
-  - `create(name, description, type_, config)` - Create a new network.
-    The name of the network is required. `description`, `type_` and `config`
+
+  - :func:`~Network.all` - Retrieve all networks.
+  - :func:`~Network.exists` - See if a profile with a name exists.
+    Returns `bool`.
+  - :func:`~Network.get` - Get a specific network, by its name.
+  - :func:`~Network.create` - Create a new network.
+    The name of the network is required. `description`, `type` and `config`
     are optional and the scope of their contents is documented in the LXD
     documentation.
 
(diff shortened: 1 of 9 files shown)


ajkavanagh:
Hi @ppkt, thanks for the PR.  It's a really great start on this feature.  However, the reason for your breakage in #302 was because you didn't detect whether the feature was available in the lxd server prior to trying to make the API call.  We've run into the issue before and added it as part of the files work; see for example: https://github.com/lxc/pylxd/blob/master/pylxd/models/container.py#L127  -- for this one, the API extension is called `network` (see: https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10networks).

This needs to be threaded into the code to throw an appropriate Exception to notify the client that the API extension isn't available.

Secondly, the code coverage bot is very picky about reducing coverage in the project (I didn't set it up!).  I agree that some tests just to try to get coverage are a bit meaningless, so if that is the case, it might be worth putting a `# NOQA` at the end of the offending lines to stop codecov spitting its dummy out!  See https://github.com/lxc/pylxd/blob/master/pylxd/models/__init__.py for an example.

Thirdly, thanks for signing off your commits!  I often forget, so it's brilliant that yours are all signed.

I'll also do an inline review.  Thanks again!

doc/source/networks.rst
```diff
@@ -11,15 +10,54 @@ Manager methods
 Networks can be queried through the following client manager
 methods:
```
ajkavanagh: Probably need a note here to say that the 'writable/savable' attributes are only available with the `network` api extension, and this is indicated by the `has_network_api_extension()`  (or whatever the function may get called!).

pylxd/models/network.py
```diff
@@ -17,18 +17,27 @@
...
     type = model.Attribute()
-    used_by = model.Attribute()
     config = model.Attribute()
-    managed = model.Attribute()
+    managed = model.Attribute(readonly=True)
+    used_by = model.Attribute(readonly=True)
+
+    @classmethod
+    def exists(cls, client, name):
+        """Determine whether a network exists."""
```
ajkavanagh: Please can we start documenting the types in the docstrings.  For this one it would be:

```
    """Ask LXD if the network called <name> exists.

    :param client: pylxd client instance
    :type client: pylxd.Client()
    :param name: the name of the network
    :type name: str
    :returns: True if the network exists
    :rtype: Boolean
    """
```

This needs to be done for each of the public facing API calls.  The reason for this is that we will be eventually transition the documentation over to  the API docstrings, and remove a lot of the 'hand written' documentation.

pylxd/models/network.py
```diff
@@ -41,14 +50,25 @@ def all(cls, client):
             networks.append(cls(client, name=name))
         return networks
 
+    @classmethod
+    def create(cls, client, name, description=None, type_=None,
+               config=None):
+        """Create a network"""
+        network = {'name': name}
```
ajkavanagh: This is where we'd do a check for the 'network' API extension and bail with an appropriate Exception.

doc/source/networks.rst
ppkt: Of course, I wasn't even aware about this function - I'll add these informations

pylxd/models/network.py
ppkt: I really like this idea - my IDE will be very happy about this. I am using [PEP 526](https://www.python.org/dev/peps/pep-0526/) but it's fairly new improvement and it's not compatible with older Python versions

pylxd/models/network.py
ppkt: I'll add this check.
BTW. Do you have any idea how to replace checking every `__init__` argument for `None` value? In my opinion we could use `**kwargs` instead and check if all keys of this dictionary are valid (using introspection for example).

pylxd/models/network.py
```diff
@@ -17,18 +17,27 @@
 class Network(model.Model):
```
ppkt: A question here - I was also thinking about adding `__str__()` / `__repr__()` methods in order to display network details in human readable form (e.g. using `marshall()`). Is there any reason why it wasn't implemented?

integration/test_networks.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_create_with_parameters(self):
+        """A network is created with provided arguments"""
+        kwargs = {
+            'name': 'eth10',
+            'config': {
+                'ipv4.address': '10.10.10.1/24',
+                'ipv4.nat': 'true',
+                'ipv6.address': 'none',
+                'ipv6.nat': 'false',
```
ppkt: I'm not quite happy about this - maybe we should accept Python data types here and cast them silently to Go before sending do LXD API?

ppkt:
@ajkavanagh one more question regarding code coverage - I'm not sure why my PR decreased it. I checked coverage on my local machine and it looks like there might be a problem in 'if' branches. Do you know where should I look to fix this?

pylxd/models/network.py
ajkavanagh: Yeah, this is a quirk of Python; there isn't a nice way to do this that also keeps readability of the function signature.  I'd just keep what you have to be honest.

pylxd/models/network.py
ajkavanagh: Yes, unfortunately, PEP526 is a Python 3.6-ism, and this library needs to maintain compatibility with Py27 for the moment.

pylxd/models/network.py
ajkavanagh: Yeah, I'd love for `__str__()` and `__repr__()` to be implemented, but I wasn't going to force it as an issue!  If you have the time to do it, then that would make this class more of an exemplar for the others to follow on with.

integration/test_networks.py
ajkavanagh: I hadn't noticed this before.  But now you've made me stop and think about it, yes, these ought to be Python types for `'true' -> True`, `'none' -> None` and `'false' -> False`.  However, I'm wondering, surely the API encoding in JSON.  */ checks */  Nope, you're right, they are stings: https://github.com/lxc/lxd/blob/master/shared/api/network.go#L25

I reckon it would be great if pylxd supported both.  In the accepting code, we could do something like:

```python
# assuming config is as above
config_ = {}
for k, v in config.items():
    if v is None:
       v = 'none'
    elif v in (True, False):
        v = str(v).lower()
```

Then, the create() call would accept both form.  Thoughts?

ajkavanagh:
yes, there seems to be something odd going on with the coverage check.  I'll re-kick the travis CI build and see if it gives new coverage.

ppkt:
@ajkavanagh thank you, I'm going to add few more features so please wait with merge. 

I'm talking about str/repr methods for Network model and converting between Go and Python data types in config.

ppkt:
@ajkavanagh of course, thanks for the tip.

ppkt:
@ajkavanagh I think my changes are done. I have few remarks:
 * I added support for str / repr functions - it's quite generic solution and can be applied to other models
 * I decided to rename one of arguments of 'Network.create' form `type_` to `type`, it shadows builtin function `type()` but it should be less confusing for users
 * there is a small decrease in coverage caused by my change in `marshall()` function and it's caused by optional attributes. It can be fixed (by removing `if` statement in _models.py:209) but I'm afraid that this might cause unexpected bugs in other places so I left it unchanged.
 * I checked LXD documentation for network API and I think we shouldn't change `config` dictionary. LXD accepts strings and using Python datatypes might be easier for user but makes code a bit complicated (especially during printing / and (de-)serializing)

PR #307, opened 2018-05-30
Update images.rst

dawncold:
`image.save()` instead of `image.update()`


ajkavanagh:
Thanks for the fix.  Please could you ensure that you commit with `git commit -s` for future PRs as (as a group defined lxc policy) all commits are supposed to be "signed off" -- see the DCO check above.

I'm merging this as it's a useful documentation fix.

Thanks!

Issue #315, opened 2018-07-05
Not able to issue migration 


gabrik:
Hi everyone,

I'm trying to Live Migrate a very simple container (an empty alpine/edge one) though two different machines
If I use the LXD cli interface the migration works pretty well
But if I try using pylxd API I got some errors, the code is pretty simple:
```
from pylxd import Client

#client_local = Client()

client_local=Client(endpoint='https://127.0.0.1:8443',cert=('lxd.crt','lxd.key'),verify=False)
client_local.authenticate('fog')
client_remote=Client(endpoint='https://192.168.1.106:8443',cert=('lxd.crt','lxd.key'),verify=False)
print('Trusted {}'.format(client_remote.trusted))
client_remote.authenticate('fog')
print('Trusted {}'.format(client_remote.trusted))

cont = client_local.containers.get('testm')
cont.migrate(client_remote,wait=True)
```
I'm using `endpoint` also for localhost otherwise I get 
```
ValueError: Cannot migrate from a local client connection
```
The error that I get is this one:
```
Traceback (most recent call last):
  File "migration.py", line 13, in <module>
    cont.migrate(client_remote,wait=True)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pylxd/models/container.py", line 267, in migrate
    self.generate_migration_data(), wait=wait)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pylxd/models/container.py", line 278, in generate_migration_data
    operation = self.client.operations.get(response.json()['operation'])
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 38, in get
    return cls(_client=client, **response.json()['metadata'])
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pylxd/models/operation.py", line 43, in __init__
    setattr(self, key, value)
AttributeError: 'Operation' object has no attribute 'description'
```

I'm misusing the API?

ajkavanagh:
Hi @gabrik 

You're not misusing the API; unfortunately the version of pylxd you are using is not compatible with the version of lxd you are using.  The `AttributeError` issue is fixed in pylxd 2.0.7 (for xenial) and 2.2.5+ for bionic.

Which version of lxd are you using (`lxd --version`) and which version of pylxd?  (`pylxd.__version__`)?

gabrik:
Hi @ajkavanagh 

I'm using lxd version 3.2 and pylxd 2.2.4 on Xenial
[...]

PR #318, opened 2018-07-09
Added example on how to use the migrate API

gabrik:
I have added the example on how to migrate a container using pylxd 
As agreed in #315
The migration fails due to an error but the example should be correct

### DIFF
--- a/doc/source/containers.rst
+++ b/doc/source/containers.rst
@@ -144,6 +144,25 @@ To get state information such as a network address.
     {'family': 'inet', 'address': '10.251.77.182', 'netmask': '24', 'scope': 'global'}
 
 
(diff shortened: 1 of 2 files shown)


ajkavanagh:
This is a great addition, thanks.  However, please could you re-do your commit with '-s' to sign the commit.  The 'lxc' project namespace requires that all commits be "signed off".

To so do, you should be able to do something similar to:

```bash
git commit -s --amend
git push -f <remote-name>
```

This will update the commit, and then update github which will then update the PR.  Relevant info for configuring git for signing off is: https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work

Thanks very much!

PR #322, opened 2018-07-25
Allow wait argument to be True/False in create_from_simplestreams and create_from_url

e1ee1e11:
In pylxd doc, the default value of `wait` argument in `create_from_simplestreams()` is `False`,
but it's forcibly set to True in [source code](https://github.com/lxc/pylxd/blob/master/pylxd/models/image.py#L155).
The problem also happens to `create_from_url()`.

As mentioned in offical REST API [document](https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10images), creating a new image is an async operation.
Therefore, I fix `wait` argument in this PR.

When `wait` is True, the function will not return until the operation is completed.
On the other hand, it will return right away when `wait` if False

### DIFF
--- a/doc/source/images.rst
+++ b/doc/source/images.rst
@@ -22,9 +22,9 @@ image:
   - `create(data, public=False, wait=True)` - Create a new image. The first
     argument is the binary data of the image itself. If the image is public,
     set `public` to `True`.
-  - `create_from_simplestreams(server, alias, public=False, auto_update=False, wait=False)` -
+  - `create_from_simplestreams(server, alias, public=False, auto_update=False, wait=True)` -
     Create an image from simplestreams.
-  - `create_from_url(url, public=False, auto_update=False, wait=False)` -
+  - `create_from_url(url, public=False, auto_update=False, wait=True)` -
     Create an image from a url.
 
 Image attributes
--- a/integration/test_images.py
+++ b/integration/test_images.py
@@ -54,6 +54,31 @@ class TestImages(IntegrationTestCase):
 
         self.assertEqual(fingerprint, image.fingerprint)
 
+    def test_create_from_simplestreams(self):
+        """An image is created from simplestreams."""
+        image = self.client.images.create_from_simplestreams(
+            'https://cloud-images.ubuntu.com/releases',
+            'trusty/amd64'
+        )
+        self.addCleanup(self.delete_image, image.fingerprint)
+
+        self.assertEqual(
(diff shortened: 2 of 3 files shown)


ajkavanagh:
Thanks for submitting the PR.  Unfortunately, the PR doesn't have any tests, which are needed before we can land it.  If you have a look at the codecov/patch and codecov/project reports above, you can see what tests are needed.  Thanks very much.

e1ee1e11:
Two questions about adding tests:
1. Where is an appropriate place to add my codes ? 
 [test_images.py](https://github.com/lxc/pylxd/blob/master/integration/test_images.py) in integration folder or [test_image.py](https://github.com/lxc/pylxd/blob/master/pylxd/tests/models/test_image.py) in tests/model folder
2. How to run tests on my machine ?

ajkavanagh:
Hi @e1ee1e11,

There are two types of tests in the pylxd library:

1. Unit tests - these are used to test individual functions do what they should do, and also help to establish the 'contract' that the function provides.  i.e. when refactoring a function, if there is no functional change, then the test should still work.  If the function of the function changes, then the test should break; thus this is important for maintenance.
2. Integration tests - these test that pylxd actually works against a controlled lxd server in a container.

The units tests are in the `pylxd/tests` directory and the integration tests are in the `integration` directory.

The unit tests for these changes should go in the `pylxd/tests/models/test_image.py` file, and any integration test changes in `integration/test_images.py`.

To run the unit tests on your machine, ensure that you have `tox` installed, and then just run `tox` from the command line in the root of the repository.  That will also run pep8 lint.

The integration tests are run using a script (actually a link) called `run_integration_tests` which relies on there being a setup version of lxd on your machine.  It creates a privileged lxd container and then runs the integration tests in that container.  There are two versions, one for Ubuntu 16.04 and one for Ubuntu 18.04 (which is now the default).

I'm on IRC on Freenode in the #lxd-dev channel; I'm tinwood, so if you need any additional info please feel free to contact me there during UK normal hours.  Otherwise, I'm happy to help here too.

(Sorry for the delay - I just got back from holiday).

e1ee1e11:
Hi @ajkavanagh ,

Great explanation about Unit tests and Integration tests!

I will find time to work on this.

Thanks.

e1ee1e11:
Hi @ajkavanagh ,

I have added two tests in `integration/test_images.py`.

Could you help to check this PR, please?

Issue #336, opened 2018-11-09
2.2.8 - Container.execute() still waits for 120 seconds in some cases.


Silentphantom62:
pylxd = 2.2.8
lxd = 3.0.2

**Steps to reproduce:**
1. Create a centos 7 container.
`lxc launch images:centos/7/amd64 my-container`

2. Run the following
```
import time

from pylxd import Client

client = Client(endpoint='https://<lxd-host>:<port>',
                verify=False, cert=('mylxd.crt', 'mylxd.key'))
client.authenticate('secret')
assert client.trusted

container = client.containers.get('my-container')

print("Started executing command -> Time:{0}".format(time.strftime('%X %x %Z')))
resp = container.execute(['useradd', '-m', 'centos', '-s', '/bin/bash'])
print("Done -> Time:{0}".format(time.strftime('%X %x %Z')))
print(resp.stdout, resp.stderr, resp.exit_code)
```

**Output:**
```
Started executing command -> Time:14:01:07 11/09/18 PST

Done -> Time:14:03:07 11/09/18 PST

(u'', u"useradd: user 'centos' already exists\n", 9)
```

Silentphantom62:
Looks like the execution is stuck in this loop waiting for all sockets to be closed.
https://github.com/lxc/pylxd/blob/master/pylxd/models/container.py#L417

Silentphantom62:
@ajkavanagh Did you get a chance to look at this?

lasizoillo:
Hi, I've created a pull request #337 to workaround the bug in ws4py. Sorry for the delay, but I've no time this days to work in it before.

ajkavanagh:
@lasizoillo hey no worries, we're all very busy!  Thanks for doing the detective work on this.  I'll do some testing with the patch today with the @Silentphantom62 use case, and if all is good, merge the patch and release 2.2.9.  Even if it's not a complete fix, it'll obviously be a step in the right direction.  Thanks again.

ajkavanagh:
So, I've done a bunch of testing with #337, which I'm going to merge.  @Silentphantom62 if you could also have a go at testing #337 that would be good.

ajkavanagh:
@Silentphantom62 does the fix in #337 fix the issue for you?

ajkavanagh:
Closing this issue as I believe #337 has fixed it.  Please re-open it if you think it's still an issue.  Thanks.

Silentphantom62:
[...]

PR #337, opened 2018-11-22
Fix race condition in 2.2.8 generating long pauses

lasizoillo:
Signed-off-by: lasizoillo <lasizoillo@gmail.com>

LXD sends TextMessage frames until the last one, then send a BinaryMessage and closes connection. If connection is closed no read events can't arrive to EPollPoller and ws4py doesn't read more data. But ws4py get into a loop in some cases waiting for streaming data to complete a message frame. There are another infinite loop waiting to ws4py to drain connection buffer.

I hope this fix resolve all corner cases. If not, I hope be more free to fix it sooner.

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -21,6 +21,7 @@ from six.moves.urllib import parse
 try:
     from ws4py.client import WebSocketBaseClient
     from ws4py.manager import WebSocketManager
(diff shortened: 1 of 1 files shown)


ajkavanagh:
Note that this may provide a fix for #336

PR #342, opened 2018-12-12
Documentation for Cluster feature

felix-engelmann:
As the cluster is now upstream, some documentation on how to use it might be helpful


doc/source/api.rst
```diff
@@ -62,3 +62,12 @@ Storage Pool
 
 .. autoclass:: pylxd.models.StoragePool
    :members:
+
+Cluster
+------------
```
ajkavanagh: Sadly, RST is a bit picky, especially in my setup; the heading underline needs to be the same length as the title above it, please.

doc/source/clustering.rst
```diff
@@ -0,0 +1,85 @@
+Clustering
+=============
```
ajkavanagh: As above.

pylxd/models/cluster.py
```diff
@@ -63,7 +63,7 @@ def get(cls, client, server_name):
 
     @classmethod
     def all(cls, client, *args):
-        """Get all certificates."""
+        """Get all cluster members."""
```
ajkavanagh: Oh, I missed this in my review on the code; good catch!

ajkavanagh:
Thanks very much for the PR - good addition to the docs, and thanks for also 'code gardening' and filling out the docstring for the container's method.

PR #343, opened 2018-12-13
Added compatibility with operations project values

felix-engelmann:
the LXD Rest api returns an operation id with the query string
`?project=default` to notify in which project the operation belongs. This
was not sufficiently cut away by `.split('/')[-1]`
but also needed a subsequent `.split('?')[0]`
    
- The Mock_lxd has been adapted to test for this feature  
- In the container.execute the operation_id is needed directly, where it
    is fixed too
    

ajkavanagh:
yes, it's almost always better to pull master, and then create a new branch for the new feature/change/bug.

PR #344, opened 2018-12-13
Added compatibility with operations project values

felix-engelmann:
the LXD Rest api returns an operation id with the query string
`?project=default` to notify in which project the operation belongs. This
was not sufficiently cut away by `.split('/')[-1]`
but also needed a subsequent `.split('?')[0]`

The Mock_lxd has been adapted to test for this feature

In the `container.execute` the `operation_id` is needed directly, where it
is fixed too

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -30,6 +30,7 @@ except ImportError:  # pragma: no cover
 from pylxd import managers
 from pylxd.exceptions import LXDAPIException
 from pylxd.models import _model as model
+from pylxd.models.operation import Operation
 
 if six.PY2:
     # Python2.7 doesn't have this natively
@@ -397,8 +398,8 @@ class Container(model.Model):
         })
 
         fds = response.json()['metadata']['metadata']['fds']
-        operation_id = response.json()['operation']\
-            .split('/')[-1].split('?')[0]
+        operation_id = \
+            Operation.extract_operation_id(response.json()['operation'])
         parsed = parse.urlparse(
             self.client.api.operations[operation_id].websocket._api_endpoint)
 
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -397,7 +397,8 @@ class Container(model.Model):
         })
 
         fds = response.json()['metadata']['metadata']['fds']
-        operation_id = response.json()['operation'].split('/')[-1]
+        operation_id = response.json()['operation']\
+            .split('/')[-1].split('?')[0]
         parsed = parse.urlparse(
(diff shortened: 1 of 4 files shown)


ajkavanagh:
This is nice; however, I do have a concern.  I'm guessing that the `?project=...` bit only exists on LXD instances that have the `projects` API extension?  In that case, this ought to check to see if that extension exists and only then try to take that bit off.  I also think that `split()` is a bit of a blunt tool here, and it would be better to try to match it and then remove it.  Possibly using `urlparse.parse_qs` might be a way of doing it in a way that pulls out the relevant parts?

```python
import urllib.parse
urllib.parse.parse_qs("https://hello.com/the/part?hello=one&project=10")

{'https://hello.com/the/part?hello': ['one'], 'project': ['10']}
```

felix-engelmann:
I agree that split is NOT the right tool. Unfortunately it is used throughout the hole project :( 
`"asdf".split('?')[0]` would still give the correct result, even if the projects extension is not enabled, but I think the urlparse option is way more pythonic

ajkavanagh:
Hey @felix-engelmann, yes, I agree split is used everywhere!  :)  I'm not one of the original maintainers, but only the most recent.  I like to try an encourage a bit of "code gardening" to see if we can improve things, and sometimes take it a bit far!!  Sorry if the request looked a bit strange in the context of the rest of the code.

My thinking went a bit like this: the original code used a `split('/')` and then grabbed the last one, which I thought was kind of okay.  But then adding the querystring params to it suddenly upped the complexity, and I reckoned "I bet urlparse has a function to do this that also errors out, etc.".  So it was the slight bump in complexity that pushed me to think - "I wonder if we can improve this little bit of code and make the library a bit more bulletproof".  I hope that's okay.

ajkavanagh:
> How do you like my static method in Operation to extract the operation id?

Yes, that's good.

> And in general, I saw that you are planning a major revision for version 3. Do you want a complete rewrite? Because then it would make little sense to put a lot of effort into this version.

It's not going to be a complete re-write, but perhaps a different set of API calls.  i.e. this version will continue for all the code that relies on it.

A lot of the 'smart' code in the library will probably be re-used, but it'll take less of a 'object orientated' approach, and more of a functional approach; i.e. removing the 'sync' function, and thus respecting that the contents of a returned 'object' are only valid at the point that it is returned.  So, no 'modify and object and save it', rather "send the changes as per the API".

However, there's not much urgency around a version 3.0, and so version 2.0 is still where it is at :)  It's more my dream of how it might be better.

PR #347, opened 2018-12-14
Avoid changing default endpoint fixes #346

### LINKED ISSUE #346: pylxd suddenly changes default endpoint
Hi all :)

The pylxd Client suddenly changes the default endpoint when a second server is installed via snap.

It stops looking to `/var/lib/lxd/unix.socket` in favour of `/var/snap/lxd/common/lxd/unix.socket`.
This is problematic as it suddenly and unexpectedly changes the server, breaking the current configuration.
It was introduced by 4f7056bfbe10916fc4109225d0cc38cf92a90e3f probably for the convenience of personal setups.

The default endpoint had to be forced for all applications using pylxd with the equivalent of the code below, to make sure that both the `lxd` command line tool and the client point to the same default and won't change without user consent when installing a parallel version just to test.

```python
import os
import pylxd

os.environ.setdefault('LXD_DIR', '/var/lib/lxd')
client = pylxd.Client()
```

We think that `/var/snap/lxd/common/lxd/unix.socket` should not take precedence over the default location.
It's prone to error, as calling the client with default arguments will then go to a completely different place.
The default `lxc` command line tool does not exhibit this behaviour. When used along, they will confusingly point to a different lxd server.
If someone installs an alternative version of lxd, it should be configured explicitly and not to take over the one it's being used without a warning. If the alternative version wants to become the default without config, it should symlink the socket to become so.

Thanks!!

jvrsantacruz:
When an alternative lxd server is installed via snap, pylxd will change the default to it, effectively
switching from the current server to a different one without the user knowing.

This was making the default value of the Client unreliable as it might change anytime without warning.

Applications relying on the default value of the endpoint for the Client had to be forced now to set `LXD_DIR` to avoid sudden changes when users get a side installation to play with.


jvrsantacruz:
Hi,

Thanks for reviewing this. I wasn't aware that the snap installation was about to be the single one. 
As to this date we used it to install alternative or modern versions and test.

I rewrote the previous commit to look for `/var/lib/lxd/unix.socket` before going for `/var/snap/lxd/common/lxd/unix.socket`. If that's more in line with what's expected of the library.

Thank you!

ajkavanagh:
Thanks for the update; I'm reviewing and testing this.  I think it's probably okay to include this functionality, but I need to check that it works after an lxd.migrate.

PR #348, opened 2018-12-15
Interactive Execute on Container

felix-engelmann:
It might be interesting to get the websockets to interact with a container, e.g. on a shell.
Therefore interactive_execute on a container returns a data and control websocket. By design of lxd, they include a secret, so no additional authentication is required to connect to them. They are then perfect to be passed to a web terminal.


ajkavanagh:
I think that this is an interesting addition, although it ought to come with a "health warning" as the caller has to do quite a bit of work to complete the command.  I think that perhaps the function should be renamed slightly (to `raw_...`) to indicate that it's low lever, and it might be useful (if possible!) to provide an example in the documentation part of the repo about how it might get used.  I'm going to add some comments to the review, but as I said, this is an interesting addition.  Let's work it through to get it landed.  Thanks!

PR #351, opened 2019-01-09
Fixed string indeces must be integers issue

nanjj:
Closes issue lxc/pylxd#350

### DIFF
--- a/pylxd/models/operation.py
+++ b/pylxd/models/operation.py
@@ -59,7 +59,9 @@ class Operation(object):
         response = self._client.api.operations[self.id].wait.get()
 
         try:
-            if response.json()['metadata']['status'] == 'Failure':
+            md = response.json()['metadata']
+            if not md: md = response.json()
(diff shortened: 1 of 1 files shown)


pylxd/models/operation.py
```diff
@@ -59,7 +59,9 @@ def wait(self):
         response = self._client.api.operations[self.id].wait.get()
 
         try:
-            if response.json()['metadata']['status'] == 'Failure':
+            md = response.json()['metadata']
+            if not md: md = response.json()
+            if md['status'] == 'Failure':
```
ajkavanagh: According to the API docs, this isn't correct.  See: https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10operationsuuid

ajkavanagh:
Note, that you also need to sign your git commits using the `-s` option.  It's a requirement set down in the lxc organisation that hosts this repository on Github.  Thanks.

Issue #301, opened 2018-04-24
UserWarning is raised for attributes from lxd that pylxd doesn't have a mapping for


shuhaowu:
Everytime I call Container.get under Ubuntu Bionic, I'm getting `/usr/local/lib/python3.6/dist-packages/pylxd-2.2.6-py3.6.egg/pylxd/models/_model.py:116: UserWarning: Attempted to set unknown attribute "location" on instance of "Container"`. I assume this is an issue because lxd 3.0 is return a location attribute in the response json?

I can't easily confirm this right now as I'm on my travel setup, but I want to document this before I forget.

ajkavanagh:
Yes, this is expected.  Due to the architecture of pylxd, it gets 'upset' when new attributes appear on objects that it is modelling.  Hence the warning.  Location is a new attribute on lxd 3.0 that isn't yet supported on pylxd 2.2.x

What really needs to happen is for pylxd to become less picky about what it will accept as an attribute, but that needs the architecture to be significantly changed (see #283 version 3.0 of pylxd), which would be a backwards incompatible change (but with legacy support).

In the interim, I'm adding the 'enhancement' tag, as it just needs the 'location' attribute added as an attribute to Container.

ajkavanagh:
So, I'm wondering; is it more useful to *not* show the warnings when attributes that pylxd doesn't have (but lxd does) are attempted to be set?  i.e. to clean up the output, but not show the warning?

ajkavanagh:
I have an alternative suggestion; suppress the warnings (about attributes) using an ENV var (default to generate warnings), and change the warning to something that indicates a later version of pylxd might include the attributes?

PR #361, opened 2019-05-03
Add PYLXD_WARNINGS env variable to be able to supress warnings

ajkavanagh:
If the LXD server that pylxd is connected to supports attributes on
objects that pylxd doesn't yet support, then a warning is issued using
the `warnings` module.  This can fill logs with annoying warnings.  So
this patch adds the ability to set an env variable PYLXD_WARNINGS to
'none' to suppress all the warnings, or to 'always' to get the existing
behaviour.  The new behavior is to issue a warning once for each
instance of an attribute that isn't known for each object.

Closes: #301
Signed-off-by: Alex Kavanagh <alex.kavanagh@canonical.com>

### DIFF
--- a/pylxd/models/_model.py
+++ b/pylxd/models/_model.py
@@ -11,6 +11,7 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+import os
 import warnings
 
 import six
@@ -83,6 +84,11 @@ class ModelType(type):
         return super(ModelType, cls).__new__(cls, name, bases, attrs)
 
 
+# Global used to record which warnings have been issues already for unknown
+# attributes.
+_seen_attribute_warnings = set()
+
+
 @six.add_metaclass(ModelType)
 class Model(object):
     """A Base LXD object model.
@@ -98,6 +104,13 @@ class Model(object):
     un-initialized attributes are read. When attributes are modified,
     the instance is marked as dirty. `save` will save the changes
     to the server.
+
+    If the LXD server sends attributes that this version of pylxd is unaware of
+    then a warning is printed.  By default the warning is issued ONCE and then
+    supressed for every subsequent attempted setting.  The warnings can be
+    completely suppressed by setting the environment variable PYLXD_WARNINGS to
+    'none', or always displayed by setting the PYLXD_WARNINGS variable to
+    'always'.
     """
     NotFound = exceptions.NotFound
     __slots__ = ['client', '__dirty__']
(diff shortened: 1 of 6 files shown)


pylxd/models/_model.py
```diff
@@ -110,6 +123,14 @@ def __init__(self, client, **kwargs):
...
                 setattr(self, key, val)
             except AttributeError:
+                global _seen_attribute_warnings
+                env = os.environ.get('PYLXD_WARNINGS', '').lower()
+                item = "{}.{}".format(self.__class__.__name__, key)
+                if env != 'always' and item in _seen_attribute_warnings:
+                    continue
+                _seen_attribute_warnings.add(item)
+                if env == 'none':
+                    continue
```
ChrisMacNaughton: Why bother adding it to the set if we're going to ignore it anyways? I'd think that it's a (tiny) waste of memory and would prefer:

```python
if env == 'none':
    continue
_seen_attribute_warnings.add(item)
```

pylxd/models/operation.py
```diff
@@ -53,6 +66,13 @@ def __init__(self, **kwargs):
             except AttributeError:
                 # ignore attributes we don't know about -- prevent breakage
                 # in the future if new attributes are added.
+                global _seen_attribute_warnings
+                env = os.environ.get('PYLXD_WARNINGS', '').lower()
+                if env != 'always' and key in _seen_attribute_warnings:
+                    continue
+                _seen_attribute_warnings.add(key)
+                if env == 'none':
+                    continue
```
ChrisMacNaughton: same comment as in _model.py

pylxd/models/_model.py
ajkavanagh: Yes, I went back and forth on this; for consistency always adding the item made sense.  The memory issue was largely irrelevant, although I did consider it.   I can change it if you like, but it doesn't change the functionality except for it the env var is changed.  I went for "changing the env var would only result in new warnings from that point" rather than "changing the env var makes it start working, but doing it again wouldn't make any difference".

pylxd/models/operation.py
ajkavanagh: as above.  thanks!

pylxd/models/_model.py
ChrisMacNaughton: Can you tell me how to change the environment variable of a running process? It seems to me that the only way to update that config would be to restart the process anyways, which means it looses any "saved state?"

ajkavanagh: Doh; indeed.  ENV vars are fixed for the run-time of the program.  I still have a nagging feeling it's more correct, from the perspective of how the variable is used; it changes the meaning of the variable.  If later, it's decided (say) that a method can be used to modify this, then, unless this is then changed, it would become a bug.

ChrisMacNaughton: Given the probably small number of unknown attributes, this doesn't bother me much

PR #370, opened 2019-07-01
changed os.path.isfile os.path.exists in client.py

tanacca:
method os.path.isfile() does not check for unix socket files

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -280,7 +280,7 @@ class Client(object):
 
         self.cert = cert
         if endpoint is not None:
-            if endpoint.startswith('/') and os.path.isfile(endpoint):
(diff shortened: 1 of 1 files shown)


ajkavanagh:
Thanks very much for the patch.  However, we need to update a few things so that this can be merged:

1. You need to sign your commit: (see the '-s' option to git commit)
2. I suspect the test is failing as the function is `exists` rather than `exist`?

tanacca:
> Thanks very much for the patch. However, we need to update a few things so that this can be merged:
> 
> 1. You need to sign your commit: (see the '-s' option to git commit)
> 2. I suspect the test is failing as the function is `exists` rather than `exist`?
 Sorry for 1. And yes definitely 2

Issue #373, opened 2019-09-08
Option for not storing stdout/stderr in container.execute


weichweich:
If you execute a command inside a container using [Container.execute](https://github.com/lxc/pylxd/blob/master/pylxd/models/container.py#L353) all output of the command is stored in a list. If the command produces to much output or is long running this becomes a memory issue.

An option to not store all output would be nice.

ajkavanagh:
This is fixed by #374.

PR #374, opened 2019-09-08
option to not store command stdout/stderr

weichweich:
this would fix issue #373

### DIFF
--- a/integration/test_containers.py
+++ b/integration/test_containers.py
@@ -155,6 +155,21 @@ class TestContainer(IntegrationTestCase):
         self.assertEqual('test\n', result.stdout)
         self.assertEqual('', result.stderr)
 
+    def test_execute_no_buffer(self):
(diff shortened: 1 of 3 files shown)


ajkavanagh:
Thanks for the patch.

However, I think there is a missed opportunity here, to drop buffering if a handler for that stream is supplied.  Then there wouldn't need to be an extra `store` parameter.

This would mean that to disable buffering (and through it away), you call `execute()` with:

```python
container.execute(..., stdout_handler=lambda _: None)
```

This would stop buffering (i.e. the `data` property would be empty) and the lambda function would just drop the data.

Also, the coverage is down as the if statements added don't have coverage; the tests would need to be altered to include them as well, please.

Thanks.

weichweich:
Hi, the idea to only buffer output if no handler is given sounds great. I implemented this version and added some tests.

PR #378, opened 2019-10-17
added image creation from public lxd server

felix-engelmann:
There is an option to pull public images from existing lxd instances, similar to simplestream.


pylxd/models/image.py
```diff
@@ -156,6 +156,32 @@ def create_from_simplestreams(cls, client, server, alias,
 
         return client.images.get(op.metadata['fingerprint'])
 
+    @classmethod
+    def create_from_image(cls, client, server, fingerprint=None, alias=None,
+                          public=False, auto_update=False, secret=None,
+                          certificate=None):
+        """Copy an image from remote lxd."""
```
ajkavanagh: Please could we have a docstring comments for the params types.  e.g.

    :param client:  the pylxd client
    :type client: pylxd.Client
    :param server: ...
    :type server: str
    ...etc.

ajkavanagh:
This is great; unfortunately, I can't merge it yet, as due to an unrelated issue, it won't pass its integration tests.  See #379 for more details.  When that's sorted, we can rebase this, and then get it merged.

PR #294, opened 2018-04-02
Update containers.rst

ilyakimavets:
Add set up config in creating container. And fix mistake in container save method

PR #384, opened 2020-02-17
Fix markdown in containers.rst

mkowalski:


mkowalski:
https://github.com/lxc/pylxd/pull/294 seems to be abandoned

ajkavanagh:
> #294 seems to be abandoned

Yes, it does, doesn't it.  I will do something about it.  Thanks for the patch!

PR #390, opened 2020-03-05
Fix simplestreams

felix-engelmann:
the image import from simplestrings has an alias parameter, which is actually the fingerprint. So this is an attempt to keep backwards compatibility and allow to actually set an alias.


sparkiegeek:
This needs rebasing, and needs unit tests writing / updating

Issue #379, opened 2019-10-21
lxd 3.18 breaks integration tests (on my machine)


ajkavanagh:
```
$ ./integration/run-integration-tests-in-lxd 
...
..
.
+ lxc config set core.https_address '[::]'
+ [[ xenial == \b\i\o\n\i\c ]]
+ echo 'Run 16.04 (xenial) integration tests'
Run 16.04 (xenial) integration tests
+ tox -e integration
integration create: /opt/pylxd/.tox/integration
integration installdeps: -r/opt/pylxd/requirements.txt, -r/opt/pylxd/test-requirements.txt
integration develop-inst: /opt/pylxd
integration installed: attrs==19.3.0,certifi==2019.9.11,cffi==1.13.0,chardet==3.0.4,coverage==4.5.4,cryptography==2.8,ddt==1.2.1,entrypoints==0.3,flake8==3.7.8,funcsigs==1.0.2,idna==2.8,mccabe==0.6.1,mock==3.0.5,mock-services==0.3.1,nose==1.3.7,pbr==5.4.3,pkg-resources==0.0.0,pycodestyle==2.5.0,pycparser==2.19,pyflakes==2.1.1,-e git+git@github.com:lxc/pylxd.git@f79a8fe7934f7498fe9c14d45019002a1127e4e8#egg=pylxd,python-dateutil==2.8.0,requests==2.22.0,requests-mock==1.1.0,requests-toolbelt==0.9.1,requests-unixsocket==0.2.0,six==1.12.0,urllib3==1.25.6,ws4py==0.5.1
integration runtests: PYTHONHASHSEED='3195312664'
integration runtests: commands[0] | nosetests integration
.S.E....SSS...............SSSSSSS......SSSSSSSSSS
======================================================================
ERROR: A command is executed on the container.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/pylxd/integration/test_containers.py", line 152, in test_execute
    result = self.container.execute(['echo', 'test'])
  File "/opt/pylxd/pylxd/models/container.py", line 440, in execute
    manager.close_all()
  File "/opt/pylxd/.tox/integration/lib/python3.5/site-packages/ws4py/manager.py", line 345, in close_all
    ws.close(code=code, reason=message)
  File "/opt/pylxd/.tox/integration/lib/python3.5/site-packages/ws4py/client/__init__.py", line 205, in close
    self._write(self.stream.close(code=code, reason=reason).single(mask=True))
[...]

PR #394, opened 2020-04-23
Fix broken pipe error on xenial integration tests

ajkavanagh:
It appears that the manager.close_all() can cause a broken pipe
error.  At this stage, that means that the pipe is closed and thus
it's actually okay to just continue and ignore the error (as per stdin)

Fixes: #379


ltrager:
Thanks for fixing that. I was able to confirm that this fixes the integration tests on Bionic.

ChrisMacNaughton:
LGTM, thanks!

PR #407, opened 2020-07-15
Added functionality to recursively pull a directory from the instance…

anneborcherding:
…  and to push an empty directory to the instance.

Co-authored-by: weichweich <14820950+weichweich@users.noreply.github.com>
Signed-off-by: Anne Borcherding <anne.borcherding@iosb.fraunhofer.de>

### DIFF
--- a/pylxd/models/instance.py
+++ b/pylxd/models/instance.py
@@ -12,12 +12,14 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 import collections
+import json
 import os
 import stat
 import time
 
 import six
 from six.moves.urllib import parse
+
 try:
     from ws4py.client import WebSocketBaseClient
     from ws4py.manager import WebSocketManager
@@ -123,6 +125,33 @@ class Instance(model.Model):
                 return
             raise LXDAPIException(response)
 
+        def put_dir(self, path, mode=None, uid=None, gid=None):
+            """Push an empty directory to the container.
+            This pushes an empty directory to the containers file system
+            named by the `filepath`.
+            :param path: The path in the container to to store the data in.
+            :type path: str
+            :param mode: The unit mode to store the file with.  The default of
+                None stores the file with the current mask of 0700, which is
+                the lxd default.
+            :type mode: Union[oct, int, str]
+            :param uid: The uid to use inside the container. Default of None
+                results in 0 (root).
+            :type uid: int
+            :param gid: The gid to use inside the container.  Default of None
+                results in 0 (root).
+            :type gid: int
+            :raises: LXDAPIException if something goes wrong
+            """
+            headers = self._resolve_headers(mode=mode, uid=uid, gid=gid)
+            headers['X-LXD-type'] = 'directory'
+            response = self._endpoint.post(
+                params={'path': path},
+                headers=headers or None)
+            if response.status_code == 200:
+                return
+            raise LXDAPIException(response)
+
         @staticmethod
         def _resolve_headers(headers=None, mode=None, uid=None, gid=None):
             if headers is None:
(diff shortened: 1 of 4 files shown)


pylxd/models/instance.py
```diff
@@ -224,6 +253,38 @@ def recursive_put(self, src, dst, mode=None, uid=None, gid=None):
...
+            :type remote_path: str
+            :param local_path: The path at which the directory will be stored.
+            :type local_path: str
+            :return:
+            :raises: LXDAPIException if an error occurs
+            """
+            response = self._endpoint.get(
+                params={'path': remote_path}, is_api=False)
+
+            print(response)
```
ajkavanagh, line 272: Is this debug?  Probably isn't wanted in the final PR.

pylxd/models/instance.py
```diff
@@ -224,6 +253,38 @@ def recursive_put(self, src, dst, mode=None, uid=None, gid=None):
...
+            if "X-LXD-type" in response.headers:
+                print("1")
+                if response.headers["X-LXD-type"] == "directory":
+                    print("2")
+                    os.mkdir(local_path)
+                    print(response.content)
```
ajkavanagh, line 278: These prints as well?

pylxd/models/instance.py
```diff
@@ -123,6 +125,33 @@ def put(self, filepath, data, mode=None, uid=None, gid=None):
                 return
             raise LXDAPIException(response)
 
+        def put_dir(self, path, mode=None, uid=None, gid=None):
+            """Push an empty directory to the container.
```
ajkavanagh, line 129: After reading what the function does, it seems more like "mk_dir" rather than "put_dir"?

pylxd/models/instance.py
```diff
@@ -123,6 +125,33 @@ def put(self, filepath, data, mode=None, uid=None, gid=None):
...
+            :param gid: The gid to use inside the container.  Default of None
+                results in 0 (root).
+            :type gid: int
+            :raises: LXDAPIException if something goes wrong
+            """
+            headers = self._resolve_headers(mode=mode, uid=uid, gid=gid)
+            headers['X-LXD-type'] = 'directory'
+            response = self._endpoint.post(
+                params={'path': path},
+                headers=headers or None)
```
ajkavanagh, line 150: `headers` will never be None here.

pylxd/models/instance.py
```diff
@@ -224,6 +253,38 @@ def recursive_put(self, src, dst, mode=None, uid=None, gid=None):
...
+            """
+            response = self._endpoint.get(
+                params={'path': remote_path}, is_api=False)
+
+            print(response)
+            if "X-LXD-type" in response.headers:
+                print("1")
+                if response.headers["X-LXD-type"] == "directory":
+                    print("2")
+                    os.mkdir(local_path)
```
ajkavanagh, line 277: What's the story on file permissions here.  At the moment they will be what ever is the umask for the parent dir?  Should they be copied/set from what is coming from the server?  If that's not the use case, then maybe put a note in the function header.

pylxd/models/instance.py
```diff
@@ -224,6 +253,38 @@ def recursive_put(self, src, dst, mode=None, uid=None, gid=None):
...
+                    print(response.content)
+                    content = json.loads(response.content)
+                    if "metadata" in content and content["metadata"]:
+                        for file in content["metadata"]:
+                            self.recursive_get(os.path.join(remote_path, file),
+                                               os.path.join(local_path, file))
+                elif response.headers["X-LXD-type"] == "file":
+                    with open(local_path, "wb") as f:
+                        f.write(response.content)
```
ajkavanagh, line 286: Again, what are the thoughts on file permissions?

pylxd/models/instance.py
anneborcherding: Thank you very much for your feedback and your thoughts! Honestly, I did not think about the possibility to set the umask to the information coming from the server. Do you think that would be the better way to go? Is there an easy way to get these information from the server that you are thinking of?

ajkavanagh: > Thank you very much for your feedback and your thoughts! Honestly, I did not think about the possibility to set the umask to the information coming from the server. Do you think that would be the better way to go? Is there an easy way to get these information from the server that you are thinking of?

Well, the REST api at https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10instancesnamefiles appears to indicate that the perms are available for the files and directories that are pulled from the LXD instance; I'm guessing that they could be used to set the permissions.  Although, there is a uid and gid, I doubt that it makes sense to use them, though, as they wouldn't necessarily translate to the host machine in any meaningful way.

I guess my question was about "is this useful in the scope of the use-case that is driving the patch"?  If not, then just perhaps add an 'TODO' comment saying something like "we considered using the X-LXD-uid, X-LXD-gid, and X-LXD-mode header information, but it was beyond the scope of this change."?

Issue #405, opened 2020-07-07
delete method of image is not documented


godfryd:
It is provided by models.Model base class but it is not present in documentation. Hard to guess that it exists.

felix-engelmann:
this issue may be closed as the update was merged

PR #417, opened 2020-11-19
Documentation for Image deletion

felix-engelmann:
This solves issue #405 as I ran into the same problem and assumed that I need to implement the delete myself.
 

ltrager:
Thanks for the fix!

PR #419, opened 2020-12-03
Blacken and isort pylxd

d0ugal:


pylxd/__init__.py
```diff
@@ -14,6 +14,6 @@
 
 import pbr.version
 
-__version__ = pbr.version.VersionInfo('pylxd').version_string()
+__version__ = pbr.version.VersionInfo("pylxd").version_string()
 
-from pylxd.client import Client,EventType  # NOQA
+from pylxd.client import Client, EventType  # NOQA
```
albertodonato, line 19: unrelated, but we could drop these NOQA and declare those imports in `__all__` (as it's the public API)

.travis.yml
```diff
@@ -1,11 +1,11 @@
 language: python
 python:
-  - "3.5"
+  - "3.8"
```
albertodonato, line 3: I think we should run against 3.6 and 3.8 for now (so we keep compatibility with bionic and focal versions of python).

We can drop 3.6 if we want to use 3.8-specific features, but for now we don't

.travis.yml
```diff
@@ -1,11 +1,11 @@
...
 env:
   matrix:
   - TOXENV=py27
```
albertodonato, line 6: We can also drop python2 compatibility

pylxd/deprecated/certificate.py
```diff
@@ -18,20 +18,22 @@
 
 
 class LXDCertificate(base.LXDBase):
-
     def certificate_list(self):
-        (state, data) = self.connection.get_object('GET', '/1.0/certificates')
-        return [certificate.split('/1.0/certificates/')[-1]
-                for certificate in data['metadata']]
+        (state, data) = self.connection.get_object("GET", "/1.0/certificates")
```
albertodonato, line 22: odd that black doesn't remove parentheses on the left side

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 1.6
-envlist = py3,py27,pypy,pep8
+envlist = py3,py27,pypy,lint
```
albertodonato, line 3: drop 2.7 from this as well?

albertodonato (changes requested):
A few comments, but looks mostly good

.travis.yml
d0ugal: Makes sense. I only made this change because black needs a more recent version than 3.5 but 3.6 will do.

.travis.yml
d0ugal: I think that should be done in a follow up, to keep this about formatting

tox.ini
d0ugal: I'll keep this for a follow up. I suspect there are other mentions on 2.7 that we can drop too

albertodonato (approved):

ajkavanagh:
Just a thought with this one; if you are dropping py27 and py35 compatibility then it probably needs a major version jump to signal that it isn't just a feature/bug fix release.

d0ugal:
@ajkavanagh Yup, good point. A major release makes sense.

PR #421, opened 2020-12-04
Remove support for Python 2.7 and test 3.6, 3.7 and 3.8

d0ugal:


doc/source/contributing.rst
```diff
@@ -78,7 +78,7 @@ Testing
 Testing pyLXD is in 3 parts:
 
 1. Conformance with `PEP 8`_, using the ``tox -e pep8`` command.
-2. Unit tests using ``tox -e py27`` and ``tox -e py3``.
+2. Unit tests using ``tox -e py3``.
```
albertodonato, line 81: we could actually `s/py3/py/` as that also matches python3 by default

doc/source/contributing.rst
```diff
@@ -126,7 +126,7 @@ Some hints on how to run the integration tests:
 1. On Ubuntu it's probably easiest to use the `Multipass`_ snap.
 2. Launch an LTS instance using ``multipass launch -n foo``
 3. Shell into the instance: ``multipass exec foo -- bash``
-4. Install tox and python2.7: ``sudo apt install python-tox python-2.7``
+4. Install tox and python3.6+: ``sudo apt install python-tox python3-dev``
```
albertodonato, line 129: I think the package is just `tox` now, but not sure from which ubuntu release.

setup.cfg
```diff
@@ -15,11 +15,10 @@ classifier =
     Operating System :: POSIX :: Linux
     Programming Language :: Python
     Programming Language :: Python :: 2
```
albertodonato, line 17: This should also go away?

requirements.txt
```diff
@@ -6,4 +6,3 @@ requests>=2.20.0
 requests-unixsocket>=0.1.5
 requests-toolbelt>=0.8.0
```
albertodonato, line 7: can we also drop dependency on `six` or is it still used somewhere?

albertodonato (changes requested):

doc/source/contributing.rst
d0ugal: I'll just include the instruction and leave folks to find the install docs if they need it.

PR #414, opened 2020-10-07
client: Add basic support for projects

stgraber:
This allows passing the project name as a string to Client and have it
automatically added to all URLs.

Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>

PR #430, opened 2020-12-07
client: Add basic support for projects

d0ugal:
This allows passing the project name as a string to Client and have it
automatically added to all URLs.

Blackened by @d0ugal for easier merging.

Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>
Signed-off-by: Dougal Matthews <dougal@dougalmatthews.com>

### DIFF
--- a/integration/test_client.py
+++ b/integration/test_client.py
@@ -35,16 +35,15 @@ class TestClient(IntegrationTestCase):
         self.assertTrue(client.trusted)
 
     def test_authenticate_with_project(self):
-
-        client = pylxd.Client(
-            "https://127.0.0.1:8443/", verify=False, project="test-project"
-        )
+        try:
+            client = pylxd.Client(
+                "https://127.0.0.1:8443/", verify=False, project="test-project"
+            )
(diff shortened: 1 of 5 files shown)


d0ugal:
This is #414 but blackened and rebased.

integration/test_client.py
```diff
@@ -24,12 +27,37 @@ class TestClient(IntegrationTestCase):
...
+        # FIXME: Once the client supports creating projects we should use that
+        #        and remove this hack.
+        os.system("lxc project create test-project")
+        self.addCleanup(os.system, "lxc project delete test-project")
+
+        client = pylxd.Client(
+            "https://127.0.0.1:8443/", verify=False, project="test-project"
+        )
+
+        self.assertFalse(client.trusted)
```
d0ugal, line 48: For some reason client.trusted in True here and in the following test. I can't figure out why, but I think there might be something wrong with this patch as it is False in the above test (which doesn't provide a project).

https://pastebin.canonical.com/p/7VT6QkSZ7M/

albertodonato (approved):

PR #433, opened 2020-12-09
configure github actions

albertodonato:


.github/workflows/unittest.yml
```diff
@@ -0,0 +1,48 @@
+name: Unit tests
+
+on:
+  push:
+    branches:
+      - main
```
sparkiegeek, line 6: pylxd still uses 'master' not 'main'

sparkiegeek (changes requested):
The names need fixing

.github/workflows/unittest.yml
albertodonato: aah, that probably explains why it's not running right now :)

sparkiegeek (approved):

PR #453, opened 2021-01-12
Docs cleanup

sparkiegeek:
Add Instance and VirtualMachine docs. Clear up Network docs.

### DIFF
--- a/pylxd/models/container.py
+++ b/pylxd/models/container.py
@@ -16,4 +16,6 @@ from pylxd.models.instance import Instance
 
 class Container(Instance):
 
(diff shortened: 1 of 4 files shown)


pylxd/models/container.py
```diff
@@ -16,4 +16,6 @@
 
 class Container(Instance):
```
albertodonato, line 18: drop the extra blank line here and in the next hunk?

albertodonato (approved):

pylxd/models/container.py
sparkiegeek: good call, fixed

PR #456, opened 2021-01-18
add tox target to typecheck via mypy

albertodonato:
The plan is to start adding type checks and verify them via mypy.

Also cleanup a bunch of mocck imports to import via unittest module.

Signed-off-by: Alberto Donato <alberto.donato@canonical.com>


sparkiegeek (approved):
:+1:

PR #464, opened 2021-02-02
Fix names of namedtuples to match variable names

sparkiegeek:
https://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html

MyPy 0.800 includes a check for the variable name of `namedtuple` to match the given name.

Signed-off-by: Adam Collard <adam.collard@canonical.com>


albertodonato (approved):
+1

PR #485, opened 2021-09-22
Enhance cluster support

cpg1111:


pylxd/models/tests/test_cluster_member.py
```diff
@@ -23,6 +23,8 @@ def test_get(self):
         member = self.client.cluster.members.get("an-member")
 
         self.assertEqual("https://10.1.1.101:8443", member.url)
+        self.assertEqual("x86_64", member.architecture)
+        self.assertEqual([], member.roles)
```
albertodonato, line 27: mind converting asserts in this test to just plain `assert`s?
We've transitioned to pytest recently, would be good to gradually move everything to `assert`

albertodonato (approved):
+1 with a nit

albertodonato:
Also you'll need a signoff line in each commit (to pass the DCO test)

pylxd/models/tests/test_cluster_member.py
cpg1111: to confirm, you mean `assert "x86_64 == member.architecture`, etc?

albertodonato: yes

albertodonato: although, `assert member.architecture == "x86_64"` seems more natural to me

PR #497, opened 2021-12-27
Fix storage type fixtures

### LINKED ISSUE #496: Type of storage volume is always custom
When I print the names and types of storage volumes I always get the custom type.
```python
pool = client.storage_pools.get(name)
volumes = pool.volumes.all()
for v in volumes:
   print(v.name, v.type)
```
Maybe the API changed at some point to singular instead of plural for the volume type.
For me the URLs that are used [here](https://github.com/lxc/pylxd/blob/1bbd68e5f685677ac306f4bc9d56eb58be03620e/pylxd/models/storage_pool.py#L354) look like this:
`/1.0/storage-pools/$POOL/volumes/container/$NAME`
instead of
`/1.0/storage-pools/$POOL/volumes/containers/$NAME`

The current swagger documentation also shows a singular type in the examples [here](https://linuxcontainers.org/lxd/api/master/#/storage/storage_pool_volumes_get).
My LXD version is: 4.0.8

atrautsch:
I changed the fixtures and storage volume type to singular.
That should be correct according to the current [swagger API example](https://linuxcontainers.org/lxd/api/master/#/storage/storage_pool_volumes_get).

However, I do not know if this was always the case.


stgraber:
Can you edit your commit message to include the `Signed-off-by: Firstname Lastname <email>` line?

You can usually do so with `git commit --amend -s` followed by a `git push --force`

stgraber:
Thanks!

PR #502, opened 2022-02-04
github-action: auto create release on tag

fliiiix:
Signed-off-by: fliiiix <hi@l33t.name>

Because of this issue [here](https://github.com/lxc/pylxd/issues/500) i thought it would be nice it the github action keeps the tags and github release in sync. 
(A bit hard to test but i used it like this in other repos)


fliiiix:
using a not deprecated action which allows to generate release notes automatically

stgraber:
Looks good, thanks!

PR #505, opened 2022-04-08
Misc fixes

simondeziel:
Most notable changes:

* use proper path to TLS client certificate/key when using the snap
* opportunistic TLS server verification attempted by looking around for a server cert with a matching name


stgraber:
```
would reformat /home/runner/work/pylxd/pylxd/integration/test_projects.py
```

PR #527, opened 2023-01-11
Fix: GitLab CI dropped support for Python 3.6 on ubuntu-latest

fliiiix:

PR #529, opened 2023-01-11
github: use 20.04 as 22.04 doesn't have python 3.6

simondeziel:


simondeziel:
Closing as #527 took care of it.

PR #532, opened 2023-02-13
Bump black to 23.1.0

sparkiegeek:
Upgrade black version to 23.1.0

Bump Ubuntu version to 22.04 from 20.04 to run with Python 3.8


albertodonato (approved):
+1

PR #533, opened 2023-02-14
Pass _APINode.session down to children classes

### LINKED ISSUE #360: _APINode is very inefficient and creates a new Requests object for every part of the path
In the `pylxd/client.py` file there is the following code in the `_APINode` class:

```python
class _APINode(object):
    """An api node object."""

    def __init__(self, api_endpoint, cert=None, verify=True, timeout=None):
        self._api_endpoint = api_endpoint
        self._timeout = timeout

        if self._api_endpoint.startswith('http+unix://'):
            self.session = requests_unixsocket.Session()
        else:
            self.session = requests.Session()
            self.session.cert = cert
            self.session.verify = verify
```

This looks okay on the face of it, except when you realise that `_APINode` is created for each part (method) of a call like:

```python
client.api.containers[name].post()
```

In this call, 3 `requests.Session()` or `requests_unixsocket.Session()` objects are created, and only the final one is used in the `post()`.  This is using up resources that simply don't need to be made AND makes testing harder as new `_APINodes` keep popping up dynamically as calls are constructed.

Going forward, the code should delay the creation of the `requests` or `requests_unixsocket` objects until the end method is called.

sparkiegeek:
Fixes #360

Allow callers to modify the session and have it persisted through the API path traversal.

Signed-off-by: Adam Collard <adam.collard@canonical.com>

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -73,24 +73,40 @@ class LXDSSLAdapter(requests.adapters.HTTPAdapter):
 
 
 class _APINode:
-    """An api node object."""
+    """An api node object.
+
+    If `session` is given, it should be a
+    `requests{,_unixsocket}.Session` which is used for issuing requests.
+
+    `cert` and `verify` options are set on the session, and should not
+    be given when `session` is given.
+    """
 
     def __init__(
-        self, api_endpoint, cert=None, verify=True, timeout=None, project=None
+        self,
+        api_endpoint,
+        cert=None,
+        verify=True,
+        timeout=None,
+        project=None,
+        session=None,
     ):
         self._api_endpoint = api_endpoint
         self._timeout = timeout
         self._project = project
 
-        if self._api_endpoint.startswith("http+unix://"):
-            self.session = requests_unixsocket.Session()
-        else:
-            self.session = requests.Session()
-            self.session.cert = cert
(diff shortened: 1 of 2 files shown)


pylxd/client.py
```diff
@@ -76,21 +76,30 @@ class _APINode:
...
 
     def __init__(
-        self, api_endpoint, cert=None, verify=True, timeout=None, project=None
+        self,
+        api_endpoint,
+        cert=None,
+        verify=True,
+        timeout=None,
+        project=None,
+        session=None,
```
albertodonato, line 85: it's probably worth adding a comment in the docstring mentioning that if the session is passed, `cert` and `verify` parameters are ignored

pylxd/tests/test_client.py
```diff
@@ -495,6 +495,12 @@ def test_session_unix_socket(self):
         node = client._APINode("http+unix://test.com")
         self.assertIsInstance(node.session, requests_unixsocket.Session)
 
+    def test_session_passed_to_chld(self):
```
albertodonato, line 498: typo, `s/chld/child/`

albertodonato (commented):
nice, +1

A couple of nits inline

albertodonato (approved):
+1

PR #535, opened 2023-02-17
Extract get_session_for_url from Client.__init__

### LINKED ISSUE #360: _APINode is very inefficient and creates a new Requests object for every part of the path

sparkiegeek:
This (really) allows callers to customise the session before it's used.

Follow-on from #533. Fixes #360

### DIFF
--- a/pylxd/tests/test_client.py
+++ b/pylxd/tests/test_client.py
@@ -458,13 +458,13 @@ class TestAPINode(TestCase):
 
     def test_getattr(self):
         """API Nodes can use object notation for nesting."""
-        node = client._APINode("http://test.com", mock.Mock())
+        node = client._APINode("http://test.com", mock.sentinel.session)
         new_node = node.test
         self.assertEqual("http://test.com/test", new_node._api_endpoint)
 
     def test_getattr_storage_pools(self):
         """API node with storage_pool should be storage-pool"""
-        node = client._APINode("http://test.com", mock.Mock())
+        node = client._APINode("http://test.com", mock.sentinel.session)
         new_node = node.test.storage_pools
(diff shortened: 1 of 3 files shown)


pylxd/tests/test_client.py
```diff
@@ -458,13 +458,13 @@ class TestAPINode(TestCase):
 
     def test_getattr(self):
         """API Nodes can use object notation for nesting."""
-        node = client._APINode("http://test.com")
+        node = client._APINode("http://test.com", mock.Mock())
```
albertodonato, line 461: please change these to `mock.sentinel.session`, so it's clear what it is

albertodonato (approved):
nice, +1

small nit inline

PR #551, opened 2023-07-11
instance(all): added passing recursion to all method to get all information in instance

### LINKED ISSUE #548: Unable to pass recursion to client.instances.all()
I have a project where I'd like to be able to get the status and the snapshots on clients when executing `client.instances.all()` instead of having to make multiple unnecessary calls to the api. Looking at the [lxd instance_get.go](https://github.com/lxc/lxd/blob/master/lxd/instances_get.go#L164) it appears that should be possible but the [pylxd all function](https://github.com/lxc/pylxd/blob/master/pylxd/models/instance.py#L294) function does not seem to support such an option

MrDaGree:
Resolves #548.

I have a project that I am working on that needs to support multi threading. Having pylxd make multiple http calls unfortunately makes my project a lot more complicated due to this. Instead of making something hacky in my project I decided to implement this to help my project and also implement more of the core LXD api.

Current implementation of `instances.all()`

```python
>>> for instance in client.instances.all():
...   print(instance.profiles)
... 
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances?recursion=0 HTTP/1.1" 200 991
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances/exampleinstance1 HTTP/1.1" 200 1910
['default']
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances/exampleinstance2 HTTP/1.1" 200 1890
['default']
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances/exampleinstance3 HTTP/1.1" 200 None
['default']
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances/exampleinstance4 HTTP/1.1" 200 1331
['default']
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances/exampleinstance5 HTTP/1.1" 200 None
['default']
```

With this PR:

```python
>>> for instance in client.instances.all(recursion=2):
...   print(instance.profiles)
... 
DEBUG:urllib3.connectionpool:https://dgreeley-test.mydomain.com:8443 "GET /1.0/instances?recursion=2 HTTP/1.1" 200 None
['default']
['default']
['default']
['default']
['default']
```

### DIFF
--- a/pylxd/models/instance.py
+++ b/pylxd/models/instance.py
@@ -291,20 +291,31 @@ class Instance(model.Model):
         return cls(client, **response.json()["metadata"])
 
     @classmethod
-    def all(cls, client):
+    def all(cls, client, recursion=0):
         """Get all instances.
 
-        Instances returned from this method will only have the name
-        set, as that is the only property returned from LXD. If more
-        information is needed, `Instance.sync` is the method call
-        that should be used.
+        This method returns an Instance array. If recursion is unset, 
+        only the name of each instance will be set and `Instance.sync` 
+        can be used to return more information. If recursion is between 
+        1-2 this method will pre-fetch additional instance attributes for 
+        all instances in the array.
         """
-        response = client.api[cls._endpoint].get()
+        response = client.api[cls._endpoint].get(params={'recursion': recursion})
 
         instances = []
-        for url in response.json()["metadata"]:
-            name = url.split("/")[-1]
-            instances.append(cls(client, name=name))
+        for instance in response.json()["metadata"]:
+            if type(instance) == dict:
+                # User specified recursion so returning all data for each instance at once
+                instance_class = cls(client, name=instance['name'])
+                for key, data in instance.items():
+                    try:
+                        setattr(instance_class, key, data)
+                    except AttributeError:
+                        pass
+                instances.append(instance_class)
+            else:
+                name = instance.split("/")[-1]
+                instances.append(cls(client, name=name))
         return instances
 
     @classmethod


pylxd/models/instance.py
```diff
@@ -291,20 +291,32 @@ def get(cls, client, name):
         return cls(client, **response.json()["metadata"])
 
     @classmethod
-    def all(cls, client):
+    def all(cls, client, **kwargs):
```
markylaing, line 294: As we only expect one optional keyword argument this might be more clear with:
```python3
def all(cls, client, recursion=0)
```

Then you can remove the call to `kwargs.get` on line 302.

pylxd/models/instance.py
```diff
@@ -291,20 +291,32 @@ def get(cls, client, name):
...
         """Get all instances.
 
         Instances returned from this method will only have the name
         set, as that is the only property returned from LXD. If more
-        information is needed, `Instance.sync` is the method call
-        that should be used.
+        information is needed, setting `recursion` between 1-2 will
+        obtain additional information
```
markylaing, line 300: I think this comment would read more clearly with:
```
This method returns an Instance array. If recursion is unset, 
only the name of each instance will be set and `Instance.sync` 
can be used to return more information. If recursion is between 
1-2 this method will pre-fetch additional instance attributes for 
all instances in the array.
```

pylxd/models/instance.py
```diff
@@ -291,20 +291,32 @@ def get(cls, client, name):
...
-            instances.append(cls(client, name=name))
+            if type(url) == dict:
+                # User specified recursion so returning all data for each instance at once
+                instance = cls(client, name=url['name'])
+                for key, data in url.items():
+                    try:
+                        setattr(instance, key, data)
+                    except AttributeError:
+                        pass
+                instances.append(instance)
```
markylaing, line 316: As we are now potentially fetching more than a list of URLs, I think the variable name here is a little misleading. I would consider renaming `url` to `resource`. Or perhaps renaming `instance` to `instance_class` and renaming `url` to `instance`.

markylaing (changes requested):
@MrDaGree thank you for adding this. Prefetching instance attributes will certainly make the client more efficient!

I've just left a couple of comments that I think will help with readability.

MrDaGree:
Thank you for the feed back! Ill get this worked in here shortly for another review

markylaing:
The changes look great thanks. Could you please squash the two commits into one though? Thanks.

MrDaGree:
Wonderful! Squashed

markylaing:
Ahh we've renamed `master` to `main` so that workflow is failing. Once #552 is merged we'll just need a quick rebase and we should be sorted.

MrDaGree:
Sounds good. I've subscribed myself so then I can get notification of that so I am on top of this

MrDaGree:
Rebased from the merge of #552

PR #553, opened 2023-07-11
Update links now that PyLXD and LXD have moved to Canonical

simondeziel:


integration/test_storage.py
```diff
@@ -93,7 +93,7 @@ def test_put(self):
         self.assertEqual(p.description, new_desc)
 
     # can't test this as patch doesn't seem to work for storage pools.
-    # Need to wait until bug: https://github.com/lxc/lxd/issues/4709
+    # Need to wait until bug: https://github.com/canonical/lxd/issues/4709
```
markylaing, line 96: Looks like this has been fixed? Maybe a follow up PR can unskip this test :smile:

pylxd/models/instance.py
```diff
@@ -853,7 +853,7 @@ def publish(self, public=False, wait=False):
         If wait=True, an Image is returned.
 
         This functionality is currently broken in LXD. Please see
-        https://github.com/lxc/lxd/issues/2201 - The implementation
+        https://github.com/canonical/lxd/issues/2201 - The implementation
```
markylaing, line 856: Looks like this one has also been fixed.

markylaing (commented):
Changes look good :+1:

Looks like there are a couple of TODOs that were waiting on LXD issues that can now be addressed.

integration/test_storage.py
simondeziel: Yep, those were meant to be addressed in a subsequent PR ;)

pylxd/models/instance.py
simondeziel: Same here

integration/test_storage.py
markylaing: Okey dokey!

markylaing (approved):
LGTM

markylaing (approved):

PR #554, opened 2023-07-13
Update code for fixed issues in LXD 

simondeziel:
This ended up fixing a small testing regression silently introduced by another PR.

>  pylxd/models/instance: only pass recursion if not 0
>  
>  This fixes `tox -e coverage`. This was not caught when introduced by
>  PR #551 because we our CI job was still only triggering on the old
>  branch name. The CI job was fixed afterward by commit 094bab4090776468.
>     
>  The decision to omit the recursion param was taken because that's easier
>  than trying to fix the mocked HTTP responses.

### DIFF
--- a/doc/source/instances.rst
+++ b/doc/source/instances.rst
@@ -211,10 +211,7 @@ us to use our own namespacing).
 A instance object (returned by `get` or `all`) has the following methods:
 
   - `rename` - rename a snapshot
-  - `publish` - create an image from a snapshot.  However, this may fail if the
-    image from the snapshot is bigger than the logical volume that is allocated
-    by lxc.  See https://github.com/canonical/lxd/issues/2201 for more details.  The solution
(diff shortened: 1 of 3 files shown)


tomponline (approved):

markylaing:
@simondeziel I'm a little confused as to why CI didn't run for the last PR, requiring you to fix a few things in this one. I had to click a button to trigger the tests, but I think only the `Commits` tests ran.

simondeziel:
@markylaing the thing is that our CI job was only triggering on push/PRs toward `master` but we had renamed it to `main` already ;)

markylaing:
> @markylaing the thing is that our CI job was only triggering on push/PRs toward `master` but we had renamed it to `main` already ;)

Ohhh that makes sense :laughing: never mind. I should have read the contributing guidelines more closely. My apologies.

PR #563, opened 2023-10-16
Add `pyproject.toml`, `requirements.txt` and `tox.ini`

simondeziel:
This modernizes the python packaging and testing bits.


simondeziel:
I got to a working state (https://github.com/simondeziel/pylxd/tree/pyproject-tox) but that required keeping the legacy `setup.py` to have the tox's `usedevelop = True` working. Since legacy bits are still needed, there isn't much point in merging this PR.

PR #570, opened 2024-01-15
Prepare `2.3.2` release

simondeziel:
```
$ python3 -m twine check --strict dist/*
Checking dist/pylxd-2.3.2-py3-none-any.whl: PASSED
Checking dist/pylxd-2.3.2.tar.gz: PASSED
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,6 +3,7 @@ name = pylxd
 version = 2.3.2a
 description = python library for LXD
 long_description = file: README.rst
+long_description_content_type = text/x-rst
 author = Paul Hummer and others (see CONTRIBUTORS.rst)
 author_email = lxd@lists.canonical.com
 home_page = https://ubuntu.com/lxd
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -3,6 +3,7 @@ name = pylxd
 version = 2.3.2a
 description = python library for LXD
 long_description = file: README.rst
+long_description_content_type = text/markdown
```
roosterfish, line 6: Since the `README.rst` is of type reStructuredText, shouldn't we proactively specify `long_description_content_type = text/x-rst` to remove the warning?

roosterfish (changes requested):

setup.cfg
simondeziel: Hmm, oops, thanks for spotting!

simondeziel: @roosterfish this error I introduced was indeed breaking the listing as we can see on the TestPyPi page: https://test.pypi.org/project/pylxd/
Thanks again!

roosterfish (approved):
LGTM!

PR #577, opened 2024-04-23
feat: add vm img support

yanksyoon:
Adds support for VM type images.


simondeziel (approved):
Thank you very much for your contribution!

PR #588, opened 2024-06-19
Skip password auth tests if not supported by LXD

simondeziel:
We don't have tests for token auth yet but that should be done as part of #589.


hamistao (approved):
LGTM!

Issue #586, opened 2024-06-12
Fix CI to actually test against various Python versions


simondeziel:
The integration tests use the `apt-get` installed Python version rather than the one set by the `setup-python` action.

PR #591, opened 2024-07-04
github: replace python versions by OS versions

simondeziel:
Due to #586, we cannot test with anything by the OS provided python version. As such, let's test with multiple OS versions and against the respective LXD version for that OS.


simondeziel:
#592 supersedes this PR.

PR #592, opened 2024-07-04
github: use canonical/setup-lxd action to setup LXD

### LINKED ISSUE #586: Fix CI to actually test against various Python versions

simondeziel:
Fixes #586, probably due to `sudo -g lxd` not preserving the `$PATH` in which
the actions/setup-python had put the python executable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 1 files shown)


hamistao (approved):
The changes LGTM. Before merging, I think this could also use a description on the PR for what it is changing since this changes the behavior of the tests and makes use of an external implementation (i.e. canonical/setup-lxd).

PR #594, opened 2024-07-05
Fix test regression on LXD 4.0

### LINKED ISSUE #590: Integration test regression on LXD 4.0
#584 introduced support for custom volume snapshot management and added additional tests. One of the additional tests now fails on LXD 4.0:

```
================================================================ FAILURES ================================================================
________________________________ TestStorageVolumeSnapshot.test_create_get_restore_delete_volume_snapshot ________________________________

self = <integration.test_storage.TestStorageVolumeSnapshot testMethod=test_create_get_restore_delete_volume_snapshot>

    def test_create_get_restore_delete_volume_snapshot(self):
        # Create pool and volume
        pool = self.create_storage_pool()
        self.addCleanup(self.delete_storage_pool, pool)
    
        volume = self.create_storage_volume(pool, "vol1")
        self.addCleanup(self.delete_storage_volume, pool, "vol1")
    
        # Create a few snapshots
>       first_snapshot = volume.snapshots.create()

integration/test_storage.py:153: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'pylxd.models.storage_pool.StorageVolumeSnapshot'>
volume = <pylxd.models.storage_pool.StorageVolume object at 0x786d535bad30>, name = None, expires_at = None

    @classmethod
    def create(cls, volume, name=None, expires_at=None):
        """Create new :class:`pylxd.models.StorageVolumeSnapshot` object from the current volume state using the given attributes.
    
        Implements POST /1.0/storage-pools/<pool>/volumes/custom/<volume_name>/snapshots
    
        :param volume: :class:`pylxd.models.StorageVolume` object that represents the target volume to take the snapshot from
        :type volume: :class:`pylxd.models.StorageVolume`
        :param name: Optional parameter. Name of the created snapshot. The snapshot will be called "snap{index}" by default.
        :type name: str
        :param expires_at: Optional parameter. Expiration time for the created snapshot in ISO 8601 format. No expiration date by default.
        :type name: str
        :returns: a storage volume snapshot if successful, raises an exception otherwise.
        :rtype: :class:`pylxd.models.StorageVolume`
        :raises: :class:`pylxd.exceptions.LXDAPIExtensionNotAvailable` if the
            'storage_api_volume_snapshots' api extension is missing.
        :raises: :class:`pylxd.exceptions.LXDAPIException` if the the operation fails.
        """
        volume.client.assert_has_api_extension("storage_api_volume_snapshots")
    
        response = volume.api.snapshots.post(
            json={"name": name, "expires_at": expires_at}
        )
    
        operation = volume.client.operations.wait_for_operation(
            response.json()["operation"]
        )
    
        # Extract the snapshot name from the response JSON in case it was not provided
        if not name:
>           name = operation.resources["storage_volume_snapshots"][0].split("/")[-1]
E           KeyError: 'storage_volume_snapshots'

pylxd/models/storage_pool.py:783: KeyError
```

An easy way to reproduce this is with `./integration/run-integration-tests-in-lxd focal`.

The issue seems to be that LXD 4.0's operation data looks like this:

```
{'storage_volumes': ['/1.0/storage_volumes/vol1']}
```

While with a newer LXD, it looks like this:

```
{'storage_volume_snapshots': ['/1.0/storage-pools/tyg1/volumes/custom/vol1/snapshots/snap0'], 'storage_volumes': ['/1.0/storage-pools/tyg1/volumes/custom/vol1']}
```

Unfortunately, I couldn't trace any specific API extension related to the addition of the `storage_volume_snapshots` key.

hamistao:
Fixes #590.
Unfortunately, the best possible solution for `create` is picking the latest snapshot name when getting the object after creation.
This is prone to error if multiple snapshots of the same volume are being taken concurrently. But since this isn't a realistic use case, these changes should be acceptable in favor of not hurting UX by returning an empty object.

### DIFF
--- a/pylxd/models/storage_pool.py
+++ b/pylxd/models/storage_pool.py
@@ -670,6 +670,37 @@ class StorageVolumeSnapshot(model.Model):
         return self.volume.api[self._endpoint][self.name]
 
     @classmethod
+    def __parse_snaspshot_json(cls, volume, snapshot_json):
+        snapshot_object = cls(volume.client, volume=volume, **snapshot_json)
+
+        # Snapshot names are namespaced in LXD, as volume-name/snapshot-name.
+        # We hide that implementation detail.
+        snapshot_object.name = snapshot_object.name.split("/")[-1]
+
+        # If reponse does not include expires_at sync the object to get that attribute.
+        if not snapshot_json.get("expires_at"):
+            snapshot_object.sync()
+
+        # Getting '0001-01-01T00:00:00Z' means that the volume does not have an expiration time set.
+        if snapshot_object.expires_at == "0001-01-01T00:00:00Z":
+            snapshot_object.expires_at = None
+
+        # Overriding default value for when created_at is not present on the response due to using LXD 4.0.
+        # Also having created_at as "0001-01-01T00:00:00Z" means this information is not available.
+        # This could be because the information was lost or the snapshot was created using an older LXD version.
+        if (
+            not volume.client.has_api_extension("storage_volumes_created_at")
+            or snapshot_json["created_at"] == "0001-01-01T00:00:00Z"
+        ):
+            snapshot_object.created_at = None
+
+        # This field is may empty so derive it from its volume.
+        if not snapshot_object.content_type:
+            snapshot_object.content_type = volume.content_type
+
+        return snapshot_object
+
+    @classmethod
     def get(cls, volume, name):
         """Get a :class:`pylxd.models.StorageVolumeSnapshot` by its name.
 
@@ -692,16 +723,7 @@ class StorageVolumeSnapshot(model.Model):
 
         response = volume.api.snapshots[name].get()
 
-        snapshot = cls(volume.client, volume=volume, **response.json()["metadata"])
-
-        # Getting '0001-01-01T00:00:00Z' means that the volume does not have an expiration time set.
-        if response.json()["metadata"]["expires_at"] == "0001-01-01T00:00:00Z":
-            snapshot.expires_at = None
-
-        # Snapshot names are namespaced in LXD, as volume-name/snapshot-name.
-        # We hide that implementation detail.
-        snapshot.name = snapshot.name.split("/")[-1]
-        return snapshot
+        return cls.__parse_snaspshot_json(volume, response.json()["metadata"])
 
     @classmethod
     def all(cls, volume, use_recursion=False):
@@ -727,19 +749,10 @@ class StorageVolumeSnapshot(model.Model):
 
         if use_recursion:
             # Using recursion so returning list of StorageVolumeSnapshot objects.
-            def parse_response_item(snapshot_json):
-                snapshot_object = cls(volume.client, volume=volume, **snapshot_json)
-                snapshot_object.content_type = volume.content_type
-                # Snapshot names are namespaced in LXD, as volume-name/snapshot-name.
-                # We hide that implementation detail.
-                snapshot_object.name = snapshot_object.name.split("/")[-1]
-
-                return snapshot_object
-
             response = volume.api.snapshots.get(params={"recursion": 1})
 
             return [
-                parse_response_item(snapshot)
+                cls.__parse_snaspshot_json(volume, snapshot)
                 for snapshot in response.json()["metadata"]
             ]
 
--- a/pylxd/models/storage_pool.py
+++ b/pylxd/models/storage_pool.py
@@ -780,7 +780,11 @@ class StorageVolumeSnapshot(model.Model):
 
         # Extract the snapshot name from the response JSON in case it was not provided
         if not name:
-            name = operation.resources["storage_volume_snapshots"][0].split("/")[-1]
+            if "storage_volume_snapshots" in operation.resources:
+                name = operation.resources["storage_volume_snapshots"][0].split("/")[-1]
+            else:
+                # If using LXD 4.0, the snapshot name isn't provided on the request response, so grab the latest snapshot name instead.
+                name = volume.snapshots.all()[-1].split("/")[-1]
 
         snapshot = volume.snapshots.get(name)
         return snapshot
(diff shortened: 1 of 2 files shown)


pylxd/models/storage_pool.py
```diff
@@ -780,7 +780,12 @@ def create(cls, volume, name=None, expires_at=None):
 
         # Extract the snapshot name from the response JSON in case it was not provided
         if not name:
-            name = operation.resources["storage_volume_snapshots"][0].split("/")[-1]
+            name = operation.resources.get("storage_volume_snapshots")
+            if name is not None:
+                name = name[0].split("/")[-1]
```
simondeziel, line 785:
```suggestion
            if "storage_volume_snapshots" in operation.resources:
                name = operation.resources["storage_volume_snapshots"][0].split("/")[-1]
```
I **think** this feels more Pythonic, WDYT?

simondeziel (commented):
Thanks for rebasing to get testing on 4.0 :)

pylxd/models/storage_pool.py
hamistao: Indeed using `in` is always very Pythonic. Guess I have become a true Go dev.

pylxd/models/storage_pool.py
```diff
@@ -698,6 +698,13 @@ def get(cls, volume, name):
         if response.json()["metadata"]["expires_at"] == "0001-01-01T00:00:00Z":
             snapshot.expires_at = None
 
+        # This field is normally empty so derive it from its volume.
+        snapshot.content_type = volume.content_type
+
+        # When dealing with older API version the response parsing can be messy, so we set those values to None.
+        if not isinstance(snapshot.created_at, str):
+            snapshot.created_at = None
```
simondeziel, line 706: There is an API extension named `storage_volumes_created_at`, which might be of interest here. It's not available on 4.0 nor 5.0 though. But maybe you can still make use of it?

simondeziel (commented):
Please better explain the rational for the `content_type` and `create_at` fiddling as I don't fully understand why they need tweaking :/

pylxd/models/storage_pool.py
hamistao: I'd prefer checking the response object because:
 - 5.0 does not include the extension but does include `created_at` on the response;
 - pyLXD does not refresh host info that often so checking for API extensions is not resilient to updates.

simondeziel: > I'd prefer checking the response object because:
> 
>     * 5.0 does not include the extension but does include `created_at` on the response;

Looks like a partial backport then. @tomponline as the original author of the feature and now the release manager, do you think we should leave things as is on LXD `stable-5.0`?

>     * pyLXD does not refresh host info that often so checking for API extensions is not resilient to updates omLXD.

pyLXD should check extension support at connection time so LXD won't change during the lifetime of that connection.

hamistao: > pyLXD should check extension support at connection time so LXD won't change during the lifetime of that connection.

Sorry for the confusion, I did mean this as an update during the lifetime of a connection. I realise this is not a problem but thought I should mention as a bonus advantage of this approach.

tomponline: > > I'd prefer checking the response object because:
> > ```
> > * 5.0 does not include the extension but does include `created_at` on the response;
> > ```
> 
> Looks like a partial backport then. @tomponline as the original author of the feature and now the release manager, do you think we should leave things as is on LXD `stable-5.0`?
> 
> > ```
> > * pyLXD does not refresh host info that often so checking for API extensions is not resilient to updates omLXD.
> > ```
> 
> pyLXD should check extension support at connection time so LXD won't change during the lifetime of that connection.

We cant add db schema changes which is why the field isnt in the db. It is in the api structs as we tend to keep them in sync as much as possible (this is what stephane told me).

But iirc the field shouldn't be back ported.

If there is an api extension then its right to check for it.

hamistao: Indeed the field is included on the response but it is always filled with the default value. I will then check for the extension for correctness.

hamistao:
@simondeziel This is ready for a review. The changes should be better explained on the commit messages, if anything is still confusing let me know.

pylxd/models/storage_pool.py
simondeziel: OK so with LXD 5.0 the extension field is not there but the `created_at` is there but always null. Make sense, thank you both for clarifying.

pylxd/models/storage_pool.py
```diff
@@ -669,6 +669,36 @@ def api(self):
...
+        # Also having created_at as "0001-01-01T00:00:00Z" means this information is not available.
+        # This could be because the information was lost or the snapshot was created using an older LXD version.
+        if (
+            not volume.client.has_api_extension("storage_volumes_created_at")
+            or snapshot_json["created_at"] == "0001-01-01T00:00:00Z"
+        ):
+            snapshot_object.created_at = None
+
+        # This field is may empty so derive it from its volume.
+        snapshot_object.content_type = volume.content_type
```
simondeziel, line 698:
```suggestion
        if not snapshot_object.content_type:
            snapshot_object.content_type = volume.content_type
```

It feels safer than blindly overwriting the value, WDYT?

simondeziel (commented):
The `content_type` handling is the last remaining concern/question I have.

pylxd/models/storage_pool.py
hamistao: I thought the same, ended up opting for this because this field should (if it doesn't it would propably be a bug) always have the same value as the volume so makes little difference. But either way is fine for me, I will push the changes.

hamistao:
@simondeziel This should be ready to merge

simondeziel (approved):

PR #595, opened 2024-07-11
integration: fix detection of explicit_trust_token API extension

simondeziel:


hamistao (approved):
Thanks for that! I was getting some errors earlier because of this bug.

PR #599, opened 2024-07-16
Add support for trust tokens

hamistao:
This updates the certificate creation to include the `trust_token` field and both the integration and migration tests. This does not update the LXD mock/unit tests and neither the broken integration tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,6 +82,7 @@ commands =
     pytest {posargs:pylxd}
 
 [testenv:integration]
+passenv = LXD_*
 commands =
     pytest integration {posargs}
 
--- a/contrib_testing/local-http-test.py
+++ b/contrib_testing/local-http-test.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import datetime
+import os
 import time
 
 import requests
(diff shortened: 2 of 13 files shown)


hamistao:
@simondeziel This is ready for a review
@ru-fu This also includes some minor changes to the docs if you want to take a look :)

setup.cfg
```diff
@@ -82,6 +82,7 @@ commands =
     pytest {posargs:pylxd}
 
 [testenv:integration]
+passenv = *
```
simondeziel, line 85: There must be a good reason for `tox` to scrape the environment variable, probably to avoid as much pollution/contamination as possible from the host running the test. As such, I'd prefer `passenv = LXD_*` or `passenv = LXD_TOKEN*`.

simondeziel (changes requested):

setup.cfg
hamistao: I like the idea of using `LXD_*`, thanks! This would include things like LXD_INSPECT, LXD_DEBUG, etc. But I think none of those can pose any problems on the tests.

simondeziel (approved):

simondeziel:
Thanks Pedro!

PR #603, opened 2024-09-09
pylxd/models/certificate: re-add password arg for backward compat

simondeziel:
Fixes https://github.com/canonical/charm-lxd/issues/168 where charm-lxd is calling certificates.create():

```python
  config: Dict[str, Union[str, bytes, List[str], bool]] = {
      "name": name,
      "password": "",
      "cert_data": cert.encode(),
  }

  client.certificates.create(**config)
```

causing:

```
  File "./src/charm.py", line 1139, in _on_https_relation_changed
    if self.lxd_trust_add(cert=cert, name=cert_name, projects=projects):
  File "./src/charm.py", line 2294, in lxd_trust_add
    client.certificates.create(**config)
TypeError: create() got an unexpected keyword argument 'password'
```

This should fix an unexpected regression introduced in commit ec1b3ee10400bf2724e7818940a6320fe2319952.

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -478,11 +478,15 @@ class Client:
         if not self.has_api_extension(name):
             raise exceptions.LXDAPIExtensionNotAvailable(name)
 
-    def authenticate(self, secret):
+    def authenticate(self, secret, use_token_auth=True):
         if self.trusted:
             return
         cert = open(self.api.session.cert[0]).read().encode("utf-8")
-        self.certificates.create(secret, cert)
+
+        if self.has_api_extension("explicit_trust_token") and use_token_auth:
+            self.certificates.create(password="", cert_data=cert, secret=secret)
+        else:
+            self.certificates.create(password=secret, cert_data=cert)
 
         # Refresh the host info
         response = self.api.get()
--- a/pylxd/models/certificate.py
+++ b/pylxd/models/certificate.py
@@ -53,12 +53,13 @@ class Certificate(model.Model):
     def create(
         cls,
         client,
-        secret,
+        password,
         cert_data,
         cert_type="client",
         name="",
         projects=None,
         restricted=False,
+        secret="",
     ):
         """Create a new certificate."""
         cert = x509.load_pem_x509_certificate(cert_data, default_backend())
@@ -68,14 +69,18 @@ class Certificate(model.Model):
         data = {
             "type": cert_type,
             "certificate": base64_cert,
+            "password": password,
             "name": name,
             "restricted": restricted,
             "projects": projects,
         }
-        if client.has_api_extension("explicit_trust_token"):
+
+        # secret/trust_token are safer than password but support for password is kept for
+        # backward compatibility
+        if client.has_api_extension("explicit_trust_token") and secret:
             data["trust_token"] = secret
-        else:
-            data["password"] = secret
+            del data["password"]
+
         response = client.api.certificates.post(json=data)
         location = response.headers["Location"]
         fingerprint = location.split("/")[-1]


hamistao:
Wouldn't it make more sense to update charm-lxd instead?

Maybe there is some limitation there that I am not aware of.

hamistao:
@simondeziel Oh I see, and that reminds me that the extension does not imply in the password being deprecated, so this would still break users using positional arguments for `certificates.create` with a LXD 5.21 that includes the extension but still supports password auth.

So I propose using `token` instead of `secret` and using it an optional parameter to indicate opting for token over password authentication. Maybe we could return an error if using the `token` arg without the extension, and also return a more informative error if not using the token with LXD 6.1, that deprecated the password option (although version checking may be undesirable). 

And maybe next release we make `token` the default option since it would break users that using password as positional parameters with 5.21.

An alternative would be using a boolean flag instead of an optional `token` parameter.

That way we allow for using password auth even with the extension and don't break any users. What do you think?

simondeziel:
The fact that this method always used `password` and recently got a `secret` argument means we have to keep those otherwise we risk breaking users.

If I missed your point, please feel free to suggest a different approach inline in the code review.

pylxd/models/certificate.py
```diff
@@ -73,9 +74,9 @@ def create(
             "projects": projects,
         }
         if client.has_api_extension("explicit_trust_token"):
-            data["trust_token"] = secret
+            data["trust_token"] = secret or password
```
hamistao, line 77: I feel like we are potentially breaking users anyway unless we dismiss the usage of positional parameters. Could you please remind me whether `pip install pylxd` grabs the latest release or uses the latest changes?

Either way my main point was allowing for password authentication even with the extension present, as I misinterpreted it as meaning that the password auth was deprecated but that is not necessarily the case, it is possible to have the extension and still use password auth. Using what I suggested but keeping the `secret` argument name as is should be better then.

If I am not doing a good job explaining I can hop on a quick call anytime :D

simondeziel: `pip install pylxd` pulls the latest release on pypi.

simondeziel: Password auth is AFAIK deprecated. Sure you can still use it on 5.21 but that's not recommended and in general when tokens are supported, password should be avoided.

hamistao: Still, don't you think denying password usage in the client even if the server allows it is not ideal?

So considering your point, why not leave `password` as an optional parameter to "force" password usage?

simondeziel:
@hamistao ready for another look when you have time, thanks!

pylxd/client.py
```diff
@@ -482,7 +482,11 @@ def authenticate(self, secret):
         if self.trusted:
             return
         cert = open(self.api.session.cert[0]).read().encode("utf-8")
-        self.certificates.create(secret, cert)
+
+        if self.has_api_extension("explicit_trust_token"):
+            self.certificates.create(password="", cert_data=cert, secret=secret)
+        else:
+            self.certificates.create(password=secret, cert_data=cert)
```
hamistao, line 489: Once again I believe we are denying passwork auth for LTS users, here I would suggest introducing a flag on the `authenticate` function (and this is an opportunity to put my new code suggestion skills to use). I like this even more because here we can convey token authentication as being the default option.

```suggestion
def authenticate(self, secret, use_token_auth=true):
        if self.trusted:
            return
        cert = open(self.api.session.cert[0]).read().encode("utf-8")

        if self.has_api_extension("explicit_trust_token") and use_trust_token:
            self.certificates.create(password="", cert_data=cert, secret=secret)
        else:
            self.certificates.create(password=secret, cert_data=cert)
```

I couldn't select one of the lines for some reason but this should get the point across.
Please let me know if you agree with this approach.

PR #611, opened 2024-10-02
Run unit tests in CI

simondeziel:


.github/workflows/tests.yml
```diff
@@ -23,14 +23,18 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install --upgrade pip tox
+        python3 -m venv venv
```
hamistao, line 26: Are these changes related to upgrading to noble?

Also I am confused as to why use `venv` here

.github/workflows/tests.yml
```diff
@@ -87,15 +87,19 @@ jobs:
     - name: Install dependencies
       run: |
         set -x
-        pip install --upgrade pip tox codecov
+        python3 -m venv venv
+        . ./venv/bin/activate
```
hamistao, line 91: I suppose the reasoning behind venv is in this fix, could you please elaborate on what the problem was?

.github/workflows/tests.yml
simondeziel: With Python 3.12 (default version shipped with Noble), using `pip` as without a `venv` gives:
```
$ lxc launch ubuntu-minimal-daily:24.04 c1
$ lxc exec c1 -- apt-get update -qq
$ lxc exec c1 -- apt-get install -y python3-pip
$ lxc exec c1 -- pip install tox
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Hence the switch to a `venv`.

hamistao: Oh now I see the relation, thanks!

hamistao (approved):
Thanks for this!
Many improvements to out CI pipeline and a significantly better developer experience for our contributors (myself included (~˘▾˘)~)

PR #618, opened 2024-12-11
pylxd/client: guard against cert=None

simondeziel:
```
>>> import pylxd
>>> c = pylxd.Client(endpoint="https://127.0.0.1:8443/", cert=None,  verify=False)
/opt/pylxd/.tox/integration/lib/python3.12/site-packages/urllib3/connectionpool.py:1099: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(
>>> c.authenticate("password")
  File "/opt/pylxd/.tox/integration/lib/python3.12/site-packages/pylxd/client.py", line 573, in authenticate
    cert = open(self.api.session.cert[0]).read().encode("utf-8")
                ~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -570,7 +570,15 @@ class Client:
     def authenticate(self, secret, use_token_auth=True):
         if self.trusted:
             return
-        cert = open(self.api.session.cert[0]).read().encode("utf-8")
+
+        if not isinstance(self.api.session.cert, tuple):
+            raise exceptions.ClientConnectionFailed("No client certificate specified")
+
+        try:
+            with open(self.api.session.cert[0]) as f:
+                cert = f.read().encode("utf-8")
+        except FileNotFoundError:
+            raise exceptions.ClientConnectionFailed("Client certificate not found")
 
         # Quirk to handle 5.21 that supports explicit trust tokens as well as
         # password auth. We need to ascertain if the provided secret is indeed a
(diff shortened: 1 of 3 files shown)


simondeziel:
@hamistao while touching the part of the code, I noticed that `Certificates.create()` is badly named as it doesn't create any cert but "onboard" an existing certificate on the LXD remote by mean of a token or a password.

hamistao:
> @hamistao while touching that part of the code, I noticed that Certificates.create() is badly named as it doesn't create any cert but "onboard" an existing certificate on the LXD remote by mean of a token or a password.

I noticed the same thing, but I don't think we can do anything about it besides maybe putting a comment to explain what the funciton does.

pylxd/client.py
```diff
@@ -570,7 +570,14 @@ def assert_has_api_extension(self, name):
...
             return
-        cert = open(self.api.session.cert[0]).read().encode("utf-8")
+
+        if not isinstance(self.api.session.cert, tuple):
+            raise exceptions.ClientConnectionFailed("No client certificate specified")
+
+        try:
+            cert = open(self.api.session.cert[0]).read().encode("utf-8")
+        except FileNotFoundError:
+            raise exceptions.ClientConnectionFailed("Client certificate not found")
```
hamistao, line 580: If the purpose is just to check if the file exists, just `open(self.api.session.cert[0])` should suffice.
Also, I think we should close the file in case it exists.

hamistao (approved):
Sorry for the sudden review, but this looks good to me!

PR #622, opened 2025-01-09
Drop `wait` argument from image create

hamistao:
This removes a useless parameter that was supposed to be removed on 2.3 release.

### DIFF
--- a/doc/source/images.rst
+++ b/doc/source/images.rst
@@ -19,7 +19,7 @@ methods:
 And create through the following methods, there's also a copy method on an
 image:
 
-  - `create(data, public=False, wait=True)` - Create a new image. The first
+  - `create(data, public=False)` - Create a new image. The first
     argument is the binary data of the image itself. If the image is public,
     set `public` to `True`.
   - `create_from_simplestreams(server, alias, public=False, auto_update=False, wait=False)` -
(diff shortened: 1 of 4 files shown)


tomponline:
@hamistao @simondeziel is this an API break?

hamistao:
@tomponline yes, it is. But it should have landed a few releases ago and since then this argument is ignored.

tomponline:
> @tomponline yes, it is. But it should have landed a few releases ago and since then this argument is ignored.

Hrm, yeah, but we recently had a separate issue with changes that broke the pylxd API, was there consideration given about users who are providing this optional value?

hamistao:
I believe @simondeziel hasn't seen this yet, so we can discuss further after he does. I think it should be safe to remove it since this parameter is being ignored for the past 8 years.

simondeziel (changes requested):
The deprecation warning was added in 2016-11 but as you pointed out @tomponline this is a breaking change. There are quite a few users of that: https://github.com/search?q=pylxd+images.create%28+language%3APython&type=code

How about you drop the first commit (c3d7c001405b69b98987e8b9c79e83289d426354) from your PR? This way, our doc and tests will stop providing this param that defaults to True anyway.
Could you rework your PR to

simondeziel (approved):

Issue #615, opened 2024-11-22
What should I do about ws4py?


dimaqq:
This is what gets installed:

```
pylxd==2.3.5
  cryptography==43.0.1
    cffi==1.17.1
      pycparser==2.22
  python-dateutil==2.9.0.post0
    six==1.16.0
  requests==2.32.3
    certifi==2024.8.30
    charset-normalizer==3.3.2
    idna==3.10
    urllib3==1.26.20
  requests-toolbelt==1.0.0
    requests==2.32.3
      certifi==2024.8.30
      charset-normalizer==3.3.2
      idna==3.10
      urllib3==1.26.20
  ws4py==0.5.1
```

And that works fine in my local env, but the Jenkins job barfs...
(I don't quite know what Jenkins does, but still?)

```
Building wheels for collected packages: juju, ws4py
  Building wheel for juju (pyproject.toml): started
  Building wheel for juju (pyproject.toml): finished with status 'done'
  Created wheel for juju: filename=juju-3.5.2.0-py3-none-any.whl size=283168 sha256=f19ece3d052a8acc76b27a8a0f0588b5ef15677554269325f877a053bf65d7e6
  Stored in directory: /home/jenkins/.cache/pip/wheels/74/32/a3/6fab14d05865e07fb4e28c16a9ff3abae6f17e145b87ce14a3
  Building wheel for ws4py (setup.py): started
  Building wheel for ws4py (setup.py): finished with status 'error'
  Running setup.py clean for ws4py
Successfully built juju
Failed to build ws4py
[31m  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [88 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib
      creating build/lib/ws4py
      copying ws4py/compat.py -> build/lib/ws4py
      copying ws4py/websocket.py -> build/lib/ws4py
      copying ws4py/async_websocket.py -> build/lib/ws4py
      copying ws4py/utf8validator.py -> build/lib/ws4py
      copying ws4py/messaging.py -> build/lib/ws4py
      copying ws4py/manager.py -> build/lib/ws4py
      copying ws4py/streaming.py -> build/lib/ws4py
      copying ws4py/framing.py -> build/lib/ws4py
      copying ws4py/__init__.py -> build/lib/ws4py
      copying ws4py/exc.py -> build/lib/ws4py
      creating build/lib/ws4py/client
[...]

Issue #585, opened 2024-06-11
Using a remote api fails in python 3.12


gchamon:
# The issue

Using python 3.12 and a remote client for lxd, I hit a problem with the `ws4py` package:

```
AttributeError                            Traceback (most recent call last)
Cell In[2], line 5
      3 client = pylxd.Client(endpoint="https://localhost:8443")
      4 instances = client.instances.all()
----> 5 instances[0].execute(["cat", "/etc/os-release"])

File ~/.local/share/virtualenvs/pylxd-yXHR9QOr/lib/python3.12/site-packages/pylxd/models/instance.py:475, in Instance.execute(self, commands, environment, encoding, decode, stdin_payload, stdin_encoding, stdout_handler, stderr_handler, user, group, cwd)
    469 stdin = _StdinWebsocket(
    470     self.client.websocket_url,
    471     payload=stdin_payload,
    472     encoding=stdin_encoding,
    473 )
    474 stdin.resource = f"{parsed.path}?secret={fds['0']}"
--> 475 stdin.connect()
    476 stdout = _CommandWebsocketClient(
    477     manager,
    478     self.client.websocket_url,
   (...)
    481     handler=stdout_handler,
    482 )
    483 stdout.resource = f"{parsed.path}?secret={fds['1']}"

File ~/.local/share/virtualenvs/pylxd-yXHR9QOr/lib/python3.12/site-packages/ws4py/client/__init__.py:214, in WebSocketBaseClient.connect(self)
    208 """
    209 Connects this websocket and starts the upgrade handshake
    210 with the remote endpoint.
    211 """
    212 if self.scheme == "wss":
    213     # default port is now 443; upgrade self.sender to send ssl
--> 214     self.sock = ssl.wrap_socket(self.sock, **self.ssl_options)
    215     self._is_secure = True
    217 self.sock.connect(self.bind_addr)

AttributeError: module 'ssl' has no attribute 'wrap_socket'
 ```
 
# Steps to reproduce

First, make sure you are using python version 3.12, then add localhost to lxc remote, changing the password to mirror your local installation:

```bash
lxc remote add localhost --accept-certificate --password an-incredibly-secure-password
```

Spin up a virtual machine, then try to communicate using the endpoint:

```python
import pylxd

client = pylxd.Client(endpoint="https://localhost:8443")
instances = client.instances.all()
instances[0].execute(["cat", "/etc/os-release"])
```

# Possible solution

[...]

PR #626, opened 2025-01-13
Switch from `ws4py` to `websockets`

hamistao:
This aims to remove `ws4py` as a required dependency for pylxd (although keeping it as an extra dependency for tests so we can test for backwards compat), and instead use the `websockets` library.
This addresses the problems reported in https://github.com/canonical/pylxd/issues/615 and https://github.com/canonical/pylxd/issues/585

### DIFF
--- a/pylxd/client.py
+++ b/pylxd/client.py
@@ -16,6 +16,7 @@ import json
 import os
 import re
 import socket
+import ssl
 from enum import Enum
 from typing import NamedTuple
 from urllib import parse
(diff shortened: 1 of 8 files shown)


hamistao:
@simondeziel To my surprise, the 3.8 tests ran indefinetely here. But before (potentially) wasting time looking for the cause, I am now thinking if it isn't time to drop support for Python 3.8 and include 3.13 in our tests. 3.8 reached EOL in October 7, 2024.

pylxd/client.py
```diff
@@ -333,7 +335,7 @@ def delete(self, *args, **kwargs):
         return response
 
 
-class _WebsocketClient(WebSocketBaseClient):
+class _WebsocketClient(ClientConnection):
```
dimaqq, line 338: Uninformed opinion:
Personally I'd prefer composition over inheritance.

pylxd/client.py
```diff
@@ -610,6 +617,58 @@ def websocket_url(self):
...
+                if len(parsed) > 1:
+                    client.protocol.wsuri.query = parsed[1]
+
+            return client
+
+        ssl_context = None
+
+        if self.api.scheme == "https" and self.cert:
+            # PROTOCOL_TLS_CLIENT does not exist on older Python versions
+            protocol = getattr(ssl, "PROTOCOL_TLS_CLIENT", ssl.PROTOCOL_TLSv1_2)
```
dimaqq, line 649: I've dug into this recently in another project and my conclusion was that PROTOCOL_TLS_CLIENT is in fact available in Python 3.8, which is the oldest Python version supported by this library.

https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT states that that's available from 3.6

the default value in `create_default_context()` became PROTOCOL_TLS_CLIENT in Python 3.10, but the constant was available long before that.

dimaqq: meanwhile PROTOCOL_TLSv1_2 has been deprecated, although not removed yet.

hamistao: Thanks for this! I guess we started using this pattern back when we supported older python versions and failed to keep track on whether the ones we support now have it.

pylxd/client.py
hamistao: I think conceptually it makes more sense to use inheritance here because the object `_WebsocketClient` represents a client, not something that includes the logic of a client. It has the capabilities of a client, no more, no less. 

Then again this is somewhat open to interpretation. Also, this fits better into the standards we set throughout pylxd code.

simondeziel:
I'll close this PR as it breaks all the exec and events tests (disabled by the last 2 commits to confirm).

auvipy:
ws4 is now maintained again. and got a new release. Also a new maintainer here

simondeziel:
> ws4 is now maintained again. and got a new release. Also a new maintainer here

Thank you @auvipy that's reassuring. Thanks for doing this often thankless job!

PR #633, opened 2025-03-18
Require GitHub actions to be pinned to their SHA1 version

simondeziel:


hamistao (approved):
Thanks for this (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

PR #637, opened 2025-03-31
pylxd/models: remove unneeded global specifier

simondeziel:


simondeziel:
> It seems those were added just for clarity, but I agree we don't need these. Thanks!

It took me a bit to understand what the issue `flake8` saw with those ;)

PR #646, opened 2025-06-09
Drop support for Python 3.8

simondeziel:
The rationals for dropping support for Python 3.8 are:

* this used to be the default Python version in Ubuntu 20.04 which is now out of standard support (EOL)
* Ubuntu releases (including 20.04 now in ESM) all use the ancient pylxd version 2.2.10 (lacking VM support) so they won't be affected
* Supporting Python 3.8 is holding back development (https://github.com/canonical/pylxd/issues/615#issuecomment-2663275517)

If accepted, this will unblock #626

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,12 +73,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.10", "3.12"]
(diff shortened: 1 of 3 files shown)


simondeziel:
GH is also just about to drop support for Python 3.8: https://github.com/actions/runner-images/issues/12034

PR #649, opened 2025-06-27
Misc

simondeziel:


integration/test_client.py
```diff
@@ -37,11 +37,6 @@ def test_authenticate(self):
         self.assertTrue(client.trusted)
 
     def test_authenticate_with_project(self):
-        if self.client.has_api_extension("explicit_trust_token"):
```
roosterfish, line 40: Do the tests run with LXD 5.21 which still has the password auth or does `pylxd` switch between the two concepts? Just out of curiosity :)

roosterfish (approved):

integration/test_client.py
simondeziel: For 5.21, only the token based auth is used by `pylxd`.

PR #732, opened 2026-04-15
Release hardening

simondeziel:
This PR hardens the pylxd release process by:

- Ensuring the `sdist` is complete and correct via `check-manifest`
- Codifying the build-and-verify steps into `tox -e release`
- Automating build attestation on tag push via a new GitHub Actions workflow

It also stop trying to upload coverage data to `codecov.io` as this wasn't working reliably and of questionable use.


simondeziel:
@tomponline could you go to https://github.com/canonical/pylxd/settings/environments and:

* create an environment named `pypi`
* add required reviewers (you, Tugba, Julian and myself please, 1 review needed)

Thanks!

.github/workflows/tests.yml
```diff
@@ -72,12 +72,6 @@ jobs:
         . ./venv/bin/activate
         tox -e coverage
 
-    - name: Upload coverage to Codecov
-      if: github.ref == 'refs/heads/main'
-      run: |
-        . ./venv/bin/activate
-        pip install --require-virtualenv codecov
-        codecov
```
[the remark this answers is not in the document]
simondeziel: That's intentional.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '[0-9]*'
```
tugbataluy: Tiny recommendation: the pattern [0-9]* matches any tag starting with a digit (e.g. 1foo, 0-test). Since the workflow only builds and attests without auto-publishing to PyPI, the blast radius is low, but a tighter pattern like '[0-9]+.[0-9]+.[0-9]+' would more precisely match SemVer release tags and avoid accidental triggers.
But what do you think?

.github/workflows/tests.yml
```diff
@@ -72,12 +72,6 @@ jobs:
         . ./venv/bin/activate
         tox -e coverage
```
tugbataluy, line 74: After removing the codecov block, there's a double blank line left before the tests. Very minor style nit.

.github/workflows/release.yml
simondeziel: I also anchored the pattern and escaped the `.`, thanks for suggesting!

tugbataluy (approved):
All previous concerns have been addressed. Thanks, Simon!

PR #740, opened 2026-04-28
Code quality

simondeziel:


simondeziel:
@tugbataluy when you get a change, this one is ready, but definitely not urgent ;)

pylxd/models/tests/test_storage.py
tugbataluy: This line still has `put_object = {"config": {"size": 1}}` while the sibling `test_patch/test_save` (the other one is in next commit)  were updated to string `"1048576" / "2097152"` to match LXD's real wire format. If the rationale for the change in `test_patch/test_save` is "match LXD behavior", then `test_put` should follow suit (and ideally also gain a body-assertion for parity).

tugbataluy:
There is a repeated pattern in `test_storage.py`. For `put`, `patch` and `delete` request:

```
matching = [r for r in self.requests_mock.request_history
            if r.method == "<METHOD>" and r.url == "<URL>"]
self.assertTrue(matching)
body = json.loads(matching[-1].text)
```
A small `_last_request(method, url)` helper on `PyLXDTestCase` (or a module-level helper next to `add_api_extension_helper`) would tighten future tests.

But it's an optional improvement, it if makes sense, maybe we could add something like that.

tugbataluy:
Very tiny typo but still worth to consider: In the commit messages I believe you tried to type "assert" not "asset", but in the comments currently I'm seeing "asset...". 
Fixing the typo doesn't add any logical improvement yet, for style and commit message accuracy I wanted to mention that.

simondeziel:
@tugbataluy thanks for the very good feedback. I've addressed all your points.

tugbataluy (approved):
Thanks Simon, to me it  looks solid and ready to be shipped 🚀 .

PR #742, opened 2026-05-12
Type checks

simondeziel:


tugbataluy:
Hi @simondeziel , this is a great, idiomatic way to initialize a set[str] in Python, thanks :sparkle: .

For even more conciseness (and zero function call overhead), you can use the empty set literal:

```
_seen_attribute_warnings: set[str] = {}
```

Both work perfectly but as a side note `{}` is slightly more performant/readable for empty sets, while `set()` shines when you later pass iterables. You can choose whichever is more suitable :)

tugbataluy (dismissed):
Thanks @simondeziel ❇️

tugbataluy:
Appearently, our linter doesn't accecpt = {} assignment, set() is also valid we can keep it as it is.

tugbataluy (approved):
Thanks again 🚀

PR #743, opened 2026-05-14
Minor fixes

simondeziel:


tugbataluy (approved):
Thanks for the detailed docstrings, now they're much easier to understand 🚀

PR #779, opened 2026-08-24
Fix opaque TypeError when publishing an instance with unsynced type/name

### LINKED ISSUE #404: Getting "TypeError: Object of type 'object' is not JSON serializable" Error When Trying to Publish Image
I have code that creates a temporary container, runs commands, then publishes the container locally as an image. 

The code is as follows:
```python
temp_container = self.mm.lxd.containers.create({
    'name': container_name, 
    'source': {
        'type': 'image', 
        'alias': template,
        'mode': 'pull'
    },
    'config': {

    },
    "devices": {
        "eth0": {
            "type": "nic",
            "nictype": "bridged",
            "parent": switch
        }
    },
}, wait=True)

temp_container.start(wait=True)
status, stdout, stderr = temp_container.execute(["/bin/sh", "-c", "sleep 5"])
for command in commands:
    self.print("Ran: " + command)
    status, stdout, stderr = temp_container.execute(["/bin/sh", "-c", command])
    if status != 0:
        self.print("Command failed!")
        self.print(stdout)
        self.print(stderr)
        temp_container.stop(wait=True)
        temp_container.delete(wait=True)

        return False

temp_container.stop(wait=True)

image = temp_container.publish(wait=True)
```

However, when I call `publish`, I get an error:
```
  File "/home/user/fakernet/lib/base_module.py", line 326, in lxd_build_image
    image = temp_container.publish(wait=True)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/pylxd/models/instance.py", line 593, in publish
    response = self.client.api.images.post(json=data)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/pylxd/client.py", line 174, in post
    response = self.session.post(self._api_endpoint, *args, **kwargs)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/requests/sessions.py", line 578, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/requests/sessions.py", line 516, in request
    prep = self.prepare_request(req)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/requests/sessions.py", line 459, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/requests/models.py", line 317, in prepare
    self.prepare_body(data, files, json)
  File "/home/user/fakernet/venv/lib/python3.6/site-packages/requests/models.py", line 468, in prepare_body
    body = complexjson.dumps(json)
  File "/usr/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'object' is not JSON serializable
```

To look at the data being sent, I modified `requests/models.py` to print out the data it's trying to convert to JSON, and noted that the `type` field is an object.
```
{'public': False, 'source': {'type': <object object at 0x7f8954fc2710>, 'name': 'hop-alpine-310-temp'}}
```

I noted that in the tests, this value is manually set in `integration/test_containers.py`:
```
# Hack to get around mocked data
self.container.type = 'container'
```
When I do this in my code, the publish completes without an error.

agu2347:
## Problem

`Instance.publish()` builds its request body directly from `self.type` and `self.name`:

```python
data = {
    "public": public,
    "source": {"type": self.type, "name": self.name},
}
```

Both are `readonly` `Attribute`s managed by the `Model` base class, which uses an internal `MISSING` sentinel (a bare `object()`) for attributes not present in the most recent `sync()` response (see `Model.sync()`: `missing_attrs = set(self.__attributes__.keys()) - set(payload.keys())`, then `setattr(self, missing_attr, MISSING)`).

If the instance's last sync response didn't include a `"type"` (or `"name"`) field, `self.type` silently evaluates to the `MISSING` sentinel rather than raising `AttributeError` -- the slot *is* set (just to the sentinel), so `Model.__getattribute__` never re-triggers `sync()`. This is not just a hypothetical: the project's own default mocked GET response for the `"an-instance"` fixture used throughout `test_instance.py` omits `"type"` entirely, which is why the existing `test_publish` has to explicitly work around it:

```python
an_instance = models.Instance(self.client, name="an-instance")
# Hack to get around mocked data
an_instance.type = "container"
```

`MISSING` then flows straight into `json=data`, and the JSON encoder fails with the opaque error reported in #404:

```
TypeError: Object of type 'object' is not JSON serializable
```

## Fix

Read `name`/`type` via the existing `_raw_attr()` helper (already used elsewhere in the codebase for exactly this "handle `MISSING` without triggering `sync()`" pattern, e.g. in `__eq__` for the `project` attribute), and raise a clear, actionable `ValueError` instead of letting the sentinel leak into the request body when either is unavailable locally.

## Testing

Added `test_publish_raises_clear_error_when_type_unknown_locally` in `pylxd/models/tests/test_instance.py`, using the default (type-less) `"an-instance"` GET mock fixture already used by the rest of the suite.

Verified:
- Against the unpatched code, this test fails with the exact reported `TypeError: Object of type object is not JSON serializable`.
- After the fix, it passes with a clear `ValueError` message.
- Reverting just the fix (keeping the test) reproduces the original crash; reapplying passes again.
- Full existing test suite: **379 passed**, no regressions -- including both the pre-existing `TestInstance.test_publish` and `TestSnapshot.test_publish` tests.

Fixes #404.


simondeziel:
@agu2347 the DCO check is failing due to missing your signed-off-by tag. Please see https://github.com/canonical/pylxd/pull/779/checks?check_run_id=97471727894 for instructions on how to add it.

tugbataluy (changes requested):
Thanks @agu2347  for digging into #404 and for the clear write-up. The diagnosis is right: `MISSING` leaks into the JSON body when the server reply has no `type`.

One thing blocks it: the new `ValueError` tells users to call `sync()`, but `sync()` is what produced the missing type, so a second call sets `MISSING` again. After this change `publish()` still cannot succeed for anyone who hits it. Also, name can never be `MISSING`, since `self.api` builds the GET URL from it.

The simpler fix is to stop depending on `self.type` at all. LXD's `POST /1.0/images` accepts the type-agnostic source type "instance" since LXD 4.0 (`lxd/images.go: case "container", "virtual-machine", "instance"`), and it is what `lxc publish` itself sends. So:

1. In `publish()`, send "type": "instance" and drop the guard.
2. In `test_publish`, replace the `an_instance.type = "container"` hack with `models.Instance.get(self.client, "an-instance")`. That makes it the regression test for #404, since the fixture has no `type`.
3. Drop `test_publish_raises_clear_error_when_type_unknown_locally`, it asserts the error that goes away.

agu2347:
@simondeziel thanks for the pointer -- amended the commit with a `Signed-off-by` trailer and updated the branch. Should pass the DCO check now.

simondeziel:
@agu2347 it still is failing and here's what it says:
> Commit sha: [02bdc69](https://github.com/canonical/pylxd/pull/779/commits/02bdc69712e78d78483a49a4a6b789f1caee683d), Author: agu2347, Committer: agu2347; Expected "agu2347 [94227848+agu2347@users.noreply.github.com](mailto:94227848+agu2347@users.noreply.github.com)", but got "agu2347 [agu2347@users.noreply.github.com](mailto:agu2347@users.noreply.github.com)".

simondeziel:
@agu2347 also, please take a look at Tugba's excellent feedback :)

agu2347:
Thanks for the review, both issues should be addressed now:

**DCO check**: the `Signed-off-by` trailer now matches this commit's actual author/committer email exactly (`agu2347 <94227848+agu2347@users.noreply.github.com>`).

**Copilot's lazy-sync concern**: agreed, reading `name`/`type` purely via `_raw_attr()` bypassed `Model`'s lazy-sync contract, so a bare, never-synced `Instance` (e.g. from `Instance.all()`) would incorrectly raise the new `ValueError` even when the server has a valid type. I've updated `publish()` so it first tries a normal attribute access (which *does* trigger `sync()` on a never-touched slot) before falling back to the `_raw_attr()` read used for the actual validation. Added `test_publish_syncs_first_when_never_synced_and_type_is_known_on_server` to cover this — verified via `git stash` that it fails with the old (addendum-less) code and passes with the fix. Full `pylxd/models/tests/test_instance.py -k publish` suite (4 tests) passes.

(Disclosure: this PR, including this update, was developed with AI assistance under human review and direction.)

simondeziel:
> Thanks for the review, both issues should be addressed now:
> 
> **DCO check**: the `Signed-off-by` trailer now matches this commit's actual author/committer email exactly (`agu2347 <94227848+agu2347@users.noreply.github.com>`).

Looks like your 2nd commit indeed pleases the DCO checker but not the first one. We prefer squashed commits when addressing PR feedback, this way it doesn't clutter the git history too much. Can you make the single commit please the DCO?

> (Disclosure: this PR, including this update, was developed with AI assistance under human review and direction.)

That's perfectly fine around here ;)

agu2347:
@simondeziel done -- squashed both commits into a single commit on the branch (force-pushed), so there's now just one commit and it carries the correct `Signed-off-by: agu2347 <94227848+agu2347@users.noreply.github.com>` trailer matching the author/committer email. Should satisfy the DCO check cleanly now.

simondeziel:
@agu2347 thanks looks better indeed. There is still the lint test failing for a minor thing though.
Also, looking at Tugba's feedback, it seems that most but not all of her points were addressed.

simondeziel:
@agu2347 hey, just a friendly ping to remind you that we are interested in the changes you bring so whenever you find to time to apply the suggested feedback, we'll be happy to get this going in. Thanks :)

